PR #5, opened 2026-06-02
Enlarge the benchmark with tlaps examples

### LINKED ISSUE #2: Enlarge the benchmark with TLAPS examples
From @tangruize : To enlarge the benchmark, https://github.com/tlaplus/Examples/ includes many examples with TLAPS proofs. There is a "TLAPS Proof" column, all specs with "TLAPS Proof :heavy_check_mark:" should be imported to the benchmark.

Saadmrp1038:
### Enlarge the benchmark with tlaplus/Examples TLAPS proofs
Closes #2

Imports every spec with a checked TLAPS proof from [tlaplus/Examples](https://github.com/tlaplus/Examples/). Generate L1 tasks into `benchmark/level1`

Following the naming convention of `<org>_<repo>_<spec>` to distinguish from existing specs.

#### Newly Added
1. 44 new imported specs, 402 new L1 tasks
2. `scripts/import_tlaps_example.py` file: imports specs from locally cloned [tlaplus/Examples](https://github.com/tlaplus/Examples/) repo. Externally linked repos are cloned into proper folder to also capture them.
3. Vendored CommunityModules: `install_deps.sh` downloads the pinned release (tag `202604221529`) into `lib/community/` (29 modules); `validate.py` and `check_proof.py` add it to the tlapm `-I` path.

#### Changes made
`generate.py`
- Added the tlapm 1.6 theorem libraries (`FiniteSetTheorems`, `FunctionTheorems`, `Folds`, the `*_proofs` variants, `RealTime`) to the stdlib set so they aren't merged/copied as local deps.
- `parse_extends` now handles `\*` line comments and multi-line `EXTENDS` that wraps onto continuation lines.
- `process_module_dir` copies the transitive deps of a copied INSTANCE dependency, not just the direct one.
- Module classification (`STDLIB_MODULES` / `COMMUNITY_MODULES` / `RESOLVABLE_MODULES`) lives here and is reused by `validate.py` and the importer.

`validate.py`
- Scoped the source-proof matcher to the task's own spec directory. Several specs share a module name (`Consensus.tla` in Paxos, PaxosHowToWinATuringAward, byzpaxos), and the old matcher could port a same-named proof from the wrong spec.

#### Validation
Run at `--jobs 4 --timeout 3600`: 392 pass / 10 fail of 402.

Each of the 10 was re-checked at `--jobs 1` (no parallelism). 7 fail for real. It seems 3 only failed under load and pass when run on their own.

Failed ones are:
- LoopInvariance -> FrontDef
- ewd998 -> NodeFact, InitiatorIsZero
- TencentPaxos -> AcceptMsgInv, Invariant, OnMessageAccInv, UpdateStateValue

With these 7 excluded we have 395 validated L1 tasks.

#### Notes
- [TPaxosWithProof.tla](https://github.com/Starydark/Tencent-Paxos-TLA/blob/master/TPaxosWithProof.tla) from an external repo and had mismatch between filename and module name. Renamed module name to `TPaxosWithProof` to match.

#### Questions
- For the 7 failing L1 task do I drop them from the `benchmark/level1` folder?
- Should I run L2 task generation?


Saadmrp1038:
@Qian-Cheng-nju lemme know if I missed anything :)
The validation part really made me crazy. I thought I was doing something wrong when I got different values for different validation runs :)
First time I ran with `--jobs 8` and got 17 failures.

Qian-Cheng-nju:
Thanks @Saadmrp1038, this is great work! Took a look and pushed two small fixes to the branch:

1. **Inline proofs**: the validator mis-ported proofs written on the same line as the statement (e.g. `LEMMA NodeFact == 0 \in Node  BY DEF Node`), which is why `NodeFact` and `InitiatorIsZero` showed up as failures. They actually pass. Fixed.
2. **CommunityModules in Docker**: the imported specs that `EXTEND` `BagsExt`/`SequencesExt`/etc. could not resolve inside the eval container. Vendored CommunityModules into the image tlapm stdlib so they resolve now.

That leaves **5 real failures**: `SumSequence_FrontDef` plus the 4 TencentPaxos lemmas (`AcceptMsgInv`, `Invariant`, `OnMessageAccInv`, `UpdateStateValue`). Their original proofs do not check under our tlapm. I think we either drop them, or fill the proofs in by hand. What do you prefer?

Saadmrp1038:
@Qian-Cheng-nju 
> I think we either drop them, or fill the proofs in by hand. What do you prefer?

You are the boss. Whatever you prefer 🫡
I only learned some TLAPS basics in the last 2 weeks :)

> the validator mis-ported proofs written on the same line as the statement

I see! Will check the failures even more carefully next time.

> could not resolve inside the eval container

Totally forgot the docker side existed.

Qian-Cheng-nju:
> That leaves 5 real failures: SumSequence_FrontDef plus the 4 TencentPaxos lemmas (AcceptMsgInv, Invariant, OnMessageAccInv, UpdateStateValue). Their original proofs do not check under our tlapm. I think we either drop them, or fill the proofs in by hand. What do you prefer?

I delete them right now.

Qian-Cheng-nju:
> Whatever you prefer 

Oh! I didn’t see that just now. 😂 Anyway, I’m planning to get the experiments running tonight first, so I removed it for now.

Issue #10, opened 2026-06-22
Single CLI entry point


Qian-Cheng-nju:
### Problem (why a new user can't run it)
- `pyproject.toml` has **no `[project.scripts]`**, so there is no `tlaps-bench` command.
- Every operation is a bare module path the user must discover:
  `python3 src/evaluator/runner.py …`, `src/common/check_proof.py`, `src/common/validate.py`,
  `src/dataset/level1/generate.py`.
- A newcomer has no obvious "where do I start"; the four entry points look unrelated.

### Solution
- Add a `[project.scripts]` console entry `tlaps-bench` that dispatches to the existing
  modules — no rewrite of their internals:
  - `tlaps-bench run`      → `evaluator/runner.py`
  - `tlaps-bench check`    → `common/check_proof.py`
  - `tlaps-bench validate` → `common/validate.py`
  - `tlaps-bench generate` → `dataset/level{1,2}/generate.py`
  - (`tlaps-bench score` is a separate task — leave a placeholder, out of scope here.)
- Works after install(#9 ) with no manual `PYTHONPATH`/venv gymnastics.

### Acceptance criteria
- [ ] `tlaps-bench --help` lists the subcommands; `tlaps-bench <sub> --help` shows the
      tool's existing flags.
- [ ] Existing flags (`--backend`, `--level`, `--filter`, `--jobs`, `--timeout`, …) reach
      the underlying tool unchanged.
- [ ] README / USAGE examples switched from `python3 src/...` to `tlaps-bench ...`.

PR #11, opened 2026-06-23
Add tlaps-bench CLI entry point dispatching to existing tools (#10)

sukanya1426:
- Register a `tlaps-bench` console command via `[project.scripts]` in `pyproject.toml` (new `src/tlaps_bench/cli.py` dispatcher).
- Five subcommands : `run`, `check`, `validate`, `generate`, `score`  that forward args to each existing module's `main()` unchanged, preserving their full flags, `--help`, and exit codes. `generate` picks level via `--level level1|level2`; `score` is a placeholder for now.
- Update README to use `tlaps-bench ...` instead of `python3 src/...`, plus a command table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,9 +148,21 @@ docker/                        Container build + isolation
 Requires Python ≥ 3.12, Java 21 (for SANY), and [uv](https://docs.astral.sh/uv/).
 
 ```bash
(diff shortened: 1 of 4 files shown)


Qian-Cheng-nju:
LGTM

PR #13, opened 2026-06-23
Add evaluation integration layer

Saadmrp1038:
### What this does

Moves agent execution into Docker containers by default along with all needed dependencies. The host just needs Docker.

The motivation is making it easy to run. Previously you needed tlapm (which errors out on glibc version mismatches), specific Java version for SANY and downloading other dependencies on the host machine. Now the docker image will handle all that.

### Design decisions

**Container-first, opt-out with `--no-container`:** Now Docker is the default. `run`, `check` and `validate` commands when run with the `tlaps-bench` cli will by default run in docker. To run localy the `--no-container` flag needs to be passed.

```bash
uv run tlaps-bench check --level 2 benchmark/level2/Consensus/Voting_ChoosableThm.tla # will run inside docker
uv run tlaps-bench check --level 2 benchmark/level2/Consensus/Voting_ChoosableThm.tla --no-container # will run on host machine
```

**Self-contained Image:** The Docker image has tlapm + SANY + checker + Node.js + Python. It's ~2.3GB. One thing I did is the agents are installed in the container at runtime instead of being pre-baked in. This helps ensure we are always using the latest versions of these tools.

**Firewall:** Moved to a standalone script. Allowed hosts are now configurable. Added IPv6 block.

**Result directory structure:** Previously everything was flat in one dir and it was hard to tell what the agent produced vs what the grader produced vs what the benchmark input was. Now the results go in dedicated folders.
```bash
$ tree
.
├── agent
│   ├── output.jsonl
│   ├── solution.tla
│   ├── stderr.txt
│   └── transcript.txt
├── grading
│   ├── agent_check.result
│   ├── check_debug.txt
│   └── check.result
├── input
│   ├── benchmark.tla
│   ├── prompt.txt
│   ├── SimpleRegular.tla
│   └── Simple.tla
└── result.json

4 directories, 12 files
```

**Grader runs in a separate container invocation:** When doing a run after the agent container exits, a fresh container runs for grading. No shared state beyond the workspace files the agent modified.

**A litellm agent with tool-calling:** Added a simple litellm-based agent that implements a full tool-calling loop (read_file, write_file, bash). This lets us test any model that litellm supports.

### Important Flags
- `--no-container`: Run locally on host instead of Docker.
- `--force-build`: Rebuild the Docker image before running. Use this after pulling new changes to `src/` or the Dockerfile, otherwise the runner will reuse the cached image. Feel free to use it on every run. If nothing changed Docker's layer cache makes it fast (<10s).

### What I tested

I tested the setup in a CloudLab VM where local setup for this benchmark does not work (outdated glibc version, missing needed Java version).

I tested all the agents claude_code, copilot, codex and litellm. I don't have subscriptions for these so I tested using the BYOK approach.

*run commands*
```bash
export ANTHROPIC_API_KEY=xxxx
uv run tlaps-bench run --backend litellm --model anthropic/claude-sonnet-4-6 --filter Simple_proof_TypeCorrect --force-build

export ANTHROPIC_API_KEY=xxxx
uv run tlaps-bench run --backend claude_code --model claude-sonnet-4-6 --filter Voting_ChoosableThm --force-build

export OPENAI_API_KEY=xxxx
uv run tlaps-bench run --backend codex --model gpt-5-nano --filter Simple_proof_TypeCorrect --force-build

export COPILOT_PROVIDER_TYPE=anthropic
export COPILOT_PROVIDER_BASE_URL=https://api.anthropic.com
export COPILOT_PROVIDER_API_KEY=xxxxx
uv run tlaps-bench run --backend copilot --model claude-sonnet-4-6 --filter Simple_proof_TypeCorrect --force-build
```

*check commands*
```bash
uv run tlaps-bench check benchmark/level1/tlaplus_examples_allocator/SimpleAllocator_proof_TypeCorrect.tla

uv run tlaps-bench check --level 2 benchmark/level2/Consensus/Voting_ChoosableThm.tla
```

*validate commands*
```bash
uv run tlaps-bench validate --filter SimpleAllocator --jobs 2
```

### Notes
1. Didn't touch the README yet since the implementation of this PR maybe be changed with review. Will do that as the last step.


tianyin:
@Saadmrp1038 feel free to merge incrementally rather than pushing a huge dragon.

Saadmrp1038:
@tianyin The whole thing needs to work together. Pushing partially can break existing build :)

Saadmrp1038:
@Qian-Cheng-nju Lemme me know how this looks or if you find any bugs. I tested the cases that came to my mind but might have missed somethings.

Qian-Cheng-nju:
Thanks @Saadmrp1038 ! this is a solid step up and the direction is clearly right.

Most of my comments will be on the **grading path**, which now runs in a container by default: a few correctness assumptions that held when grading ran on the host no longer do. I'll left them inline, and I'd like those resolved before merge.

BTW, this touches `check_proof.py` / `runner.py` / `gates.py`, which overlap in-flight cheat-checker work on another branch, I'll reconcile that side once this lands.

src/common/check_proof.py
```diff
@@ -499,13 +560,27 @@ def main():
...
+    )
     args = parser.parse_args()
 
     filepath = os.path.abspath(args.file)
     if not os.path.isfile(filepath):
         print(f"ERROR: File not found: {filepath}")
         sys.exit(3)
 
+    # Default: run in Docker. --no-container runs locally.
+    if not args.no_container:
```
Qian-Cheng-nju: The agent's own self-check can't run in-container. The prompts have the agent iterate with `check_proof_bin {file} --level N` (no `--no-container`), but that now defaults to Docker, and the image has no docker CLI/socket, so it errors out and the agent loses its PASS/CHEATING feedback. Fix: default the baked-in binary to `--no-container` when it's already inside the image (e.g. a `TLAPS_IN_CONTAINER=1` env set in the Dockerfile).

src/evaluator/runner.py
```diff
@@ -436,194 +442,318 @@ def run_single_benchmark(item: WorkItem):
...
+
+    # Use container path for checker binary (not host path)
+    old_binary = level._checker_binary
+    level._checker_binary = "/usr/local/bin/check_proof_bin"
+    check_cmd = level.checker_command(
+        "/workspace",
+        basename,
+        "/results/check.result",
+        item.check_timeout,
+        benchmark_dir="/workspace",
```
Qian-Cheng-nju: This hands the grader the agent's modified `/workspace` as the cheat-check baseline (the local grader passes the read-only source). Since `bench = benchmark_dir or git_bench` in `check_proof.py`, `/workspace` overrides the pristine git baseline, so the diff-based gates (statement / extra-axiom / deps / admitted-set) compare the solution against itself and never fire. Fix: pass `benchmark_dir=None` and set `safe.directory=/workspace` (as `_run_in_container` already does) so the baseline is the seeded pristine git commit.

src/common/check_proof.py
Saadmrp1038: Wow. Just checked a couple of my agent traces and the agent figured out the usage of `--no-container` on its own. First made a call without `--no-container`. It failed. Then it checked the `--help` manual and ran with `--no-container` flag.

Qian-Cheng-nju: That's interesting! That said, I think we still need to fix that.

src/evaluator/runner.py
Saadmrp1038: I see what happened here :)
Fixed

src/common/container.py
```diff
@@ -0,0 +1,310 @@
...
+        args.append(config.image)
+        return args, cid_file
+
+    def build_composite_command(self, cmd: list[str], install_script: str | None = None) -> str:
+        """Build shell command: install script → firewall → agent command."""
+        agent_cmd = " ".join(shlex.quote(c) for c in cmd)
+        parts = []
+        if install_script:
+            parts.append(f"/opt/install-scripts/{install_script} >&2")
+        parts.append("/opt/firewall.sh >&2 || true")
```
Qian-Cheng-nju: `firewall.sh` runs under `set -e`, so a partial failure (a `dig` or `iptables` call) makes it exit before the final `-j DROP`, and `|| true` swallows that — the agent can end up with the firewall only partially applied (or open), indistinguishable from success. Fix: drop `|| true`, and have `firewall.sh` confirm at least one ACCEPT rule was installed per host and exit non-zero otherwise.

src/common/container.py
```diff
@@ -0,0 +1,310 @@
...
+        # tlapm is baked into the image at /opt/tlapm
+
+        # Credential directory mounts (read-only except codex which needs write for cache)
+        aws_dir = Path.home() / ".aws"
+        if aws_dir.is_dir():
+            args.extend(["-v", f"{aws_dir.resolve()}:/root/.aws:ro"])
+
+        codex_dir = Path.home() / ".codex"
+        if codex_dir.is_dir():
+            args.extend(["-v", f"{codex_dir.resolve()}:/root/.codex"])
```
Qian-Cheng-nju: `~/.codex` and `~/.copilot` are mounted read-write into a container that runs as root with an agent that has a `bash` tool (only `~/.aws` is `:ro`), so a misbehaving agent can read or overwrite host credentials. Fix: mount `:ro` (or only the cache subdir that needs writing), and only mount the dir for the selected backend.

Saadmrp1038: The no `ro` is intentional.  Without write access `codex` cannot write down agent trace (empty output.jsonl). Haven't checked with all of them though.

src/evaluator/runner.py
```diff
@@ -436,194 +442,321 @@ def run_single_benchmark(item: WorkItem):
...
+        container_run = runner.run(config, cmd, stdin_data=prompt)
+        proc = container_run.proc
+        assert proc.stdout is not None  # Popen created with stdout=PIPE
+
+        # Stream stdout to file in real-time (and stderr separately)
+        with open(agent_jsonl, "w") as jsonl_f:
+            deadline = (time.time() + timeout) if timeout else None
+            while True:
+                # Poll with 5s timeout so deadline is checked even if agent hangs
+                ready, _, _ = select.select([proc.stdout], [], [], 5.0)
```
Qian-Cheng-nju, line 555: The loop only drains `proc.stdout`; `stderr` is read after it finishes. If the agent writes more than the pipe buffer (~64KB) to stderr, that write blocks and the run can hang until the timeout. Better to drain stderr in the same loop, consistently with stdout.

src/common/container.py
Saadmrp1038: It seems `copilot` also needs rw access. Without it fails to write session state.

Qian-Cheng-nju: Got it. I think the cleanest fix is to mount a throwaway copy: copy ~/.codex into a per-run temp dir and mount that rw. Although I think AI most likely won't cause any harm, it would still be better to be more standardized. What do you think?

src/evaluator/backends/base.py
```diff
@@ -4,39 +4,65 @@
...
+        "api.githubcopilot.com",
+        "api.business.githubcopilot.com",
+        "api.enterprise.githubcopilot.com",
+    ]
+    + BEDROCK_HOSTS
+    + VERTEX_HOSTS
+)
+
+
+def detect_firewall_hosts(model: str) -> list[str]:
```
Qian-Cheng-nju, line 57: `detect_firewall_hosts` ignores `model`, so every backend can reach all provider hosts, and `container.py` forwards all API keys on the host. So an agent under test could just call a different provider's model to solve the task. If isolation is the goal, limit each backend to its own host + key. (Minor: `model` is unused here, and the comment above has a typo: "internal" → "internet".)

src/common/container.py
```diff
@@ -0,0 +1,312 @@
...
+            f"--cidfile={cid_file}",
+        ]
+
+        if config.cpus:
+            args.append(f"--cpus={config.cpus}")
+        if config.memory:
+            args.append(f"--memory={config.memory}")
+
+        if config.cap_net_admin and config.firewall_hosts:
+            args.append("--cap-add=NET_ADMIN")
```
munimthahmid, line 230: This grants NET_ADMIN to the root container running the agent, so the agent can execute iptables -F OUTPUT and restore unrestricted network access.I think this defeats the anti-cheating isolation

src/evaluator/backends/base.py
Saadmrp1038: The initial implementation of this was model based. But later it occured to me it's really hard to detect the provider of a model in some cases. If someone runs a claude/openai model we don't really know if it's hosted from antrhopic/openai, github, azure or from bedrock. Things become even more murky for opensource models which can have many many different providers (they are not even added yet). That's why I just bunched all providers together.

> So an agent under test could just call a different provider's model to solve the task.

This seems highly unlikely doesn't it? :)

---
If there's another type of implementation in mind please suggest :)

src/common/container.py
Saadmrp1038: This could work. Let me try it out.

Saadmrp1038: Seems to be working!

src/common/container.py
Saadmrp1038: Ooo. Gave a fix to drop NET_ADMIN after the firewall is set up using [capsh](https://man7.org/linux/man-pages/man1/capsh.1.html). So, the agent shouldn't be able to flush the iptables rules even as root.

src/evaluator/runner.py
```diff
@@ -436,194 +443,347 @@ def run_single_benchmark(item: WorkItem):
...
+
+    # Use container path for checker binary (not host path)
+    old_binary = level._checker_binary
+    level._checker_binary = "/usr/local/bin/check_proof_bin"
+    check_cmd = level.checker_command(
+        "/workspace",
+        basename,
+        "/results/check.result",
+        item.check_timeout,
+        benchmark_dir=None,  # let git-root reconstruction provide the pristine baseline
```
munimthahmid: Here, the grader reconstructs the baseline from the workspace's git repo. But the agent has write access to that workspace, so it can `git commit --amend` to rewrite the initial benchmark commit itself ( tamper with the very baseline it's graded against).

src/evaluator/backends/claude_code.py
```diff
@@ -6,30 +6,46 @@
...
+        "WebFetch",
+        "NotebookEdit",
+        "NotebookRead",
+        "TodoRead",
+        "TodoWrite",
+        "Agent",
+        "Skill",
+        "SlashCommand",
+        "Task",
+        "WebSearch",
```
munimthahmid: WebSearch` runs server-side over `api.anthropic.com` (which the firewall must allow), so it bypasses network isolation, and the agent can look up online sources. We can probably remove this.

src/evaluator/backends/base.py
Qian-Cheng-nju: > This seems highly unlikely doesn't it? :)

Yeah, I agree that this should not occur (I have previously seen such issues in some places, for example: https://arxiv.org/html/2605.15229v1), but it should not be a problem for us I guess.

src/common/container.py
Qian-Cheng-nju: `|| true` lets the agent run even when `firewall.sh` failed. Since the OUTPUT default policy is ACCEPT and the `-j DROP` rule is only added at the very end, any partial failure leaves the network fully open silently. Fix: drop `|| true` so the agent doesn't start when the firewall didn't come up.

src/evaluator/runner.py
```diff
@@ -436,194 +443,348 @@ def run_single_benchmark(item: WorkItem):
...
+
+    # Use container path for checker binary (not host path)
+    old_binary = level._checker_binary
+    level._checker_binary = "/usr/local/bin/check_proof_bin"
+    check_cmd = level.checker_command(
+        "/workspace",
+        basename,
+        "/results/check.result",
+        item.check_timeout,
+        benchmark_dir="/benchmark",  # tamper-proof read-only mount
```
Qian-Cheng-nju, line 702: This `/benchmark:ro` isn't actually used as the baseline. `build_context` reads the baseline from `solution_dir/benchmark.tla` (`tlacheck/context.py`), which `check_proof.py` fills from the **workspace git root commit**, not from `/benchmark`. Since the agent owns `/workspace` and has a shell, `git add -A && git commit --amend` rewrites that root commit to its own cheated version, then the statement / extra-axiom / admitted diff gates compare the solution against itself and never fire. We've previously observed agents deliberately hacking the cheat checker, so this one really needs fixing. Fix: have `build_context` read the baseline `<target>.tla` from `benchmark_dir` (this `/benchmark` mount), not from `solution_dir/benchmark.tla`, so the baseline is outside the agent-writable workspace.

src/common/container.py
Saadmrp1038: Dropped `|| true`. Also, I noticed I was just counting ACCEPT and not per-host count. Added that. Can u check now?

Qian-Cheng-nju: LGTM. Thank you!

src/evaluator/runner.py
Saadmrp1038: Reading baseline from `benchmark_dir` when available. We should setup some tests for these cases when there's time :)

Qian-Cheng-nju: LGTM, Thank you! I'll add a regression case on the cheat checker side: amend the seeded root commit, then confirm the gate still fires off

src/common/container.py
```diff
@@ -0,0 +1,328 @@
...
+        if config.benchmark_dir:
+            args.extend(["-v", f"{config.benchmark_dir}:/benchmark:ro"])
+
+        # tlapm is baked into the image at /opt/tlapm
+
+        # Credential directory mounts. Copy to throwaway tempdirs so agent
+        # cannot modify host credentials. Cleaned up after run.
+        self._credential_tmps: list[str] = []
+
+        for name, mount_path in [("aws", "/root/.aws"), ("codex", "/root/.codex"), ("copilot", "/root/.copilot")]:
```
munimthahmid: This still copies and mounts the entire ~/.codex directory for every container run, including non-Codex backends and grader/TLAPM containers. A tested agent can read prior Codex sessions/logs/config and potentially recover previous benchmark IDs or proofs. I think one fix can be to mount credentials only for the selected backend, and for Codex use a per-run home containing only the minimal auth file, e.g. auth.json, instead of copying the full directory.

Saadmrp1038: We are mounting a copy always. So technically if the benchmark is ran using container always the host .codex folder won't have session to look at. The copy gets thrown away after each run. The session-leakage concern applies if you mix --no-container runs (which write directly to host) with container run.

Saadmrp1038: To still be on the safe side we can clean the existing session/logs from the mounted copy (as u said minimal auth file)

Saadmrp1038: @munimthahmid Can you please check now? Unfortunately, I don't have sub for any of these (claudecode/codex/copilot). So, I can't really do proper testing. If you find anything wrong just give a push to this branch since I can't really test these properly anyway.

munimthahmid: I tested with Codex subscription auth. The minimal Codex credential copy works: the container only saw auth.json, and codex login status worked fine.

One remaining issue: firewalled codex exec times out with the current allowlist because it only includes api.openai.com for OpenAI/Codex. I believe we also need to allow chatgpt.com for Codex ChatGPT/subscription auth, since that path uses https://chatgpt.com/backend-api/ per the OpenAI docs.

munimthahmid: Also tested with claude code subscription. It's working fine

Saadmrp1038: Thanks!

Saadmrp1038: Added `chatgpt.com` and `auth.openai.com`.

PR #15, opened 2026-06-24
Cheat checker: framework

Qian-Cheng-nju:
# Cheat checker: framework

The Lean theorem-prover community ran into the same problem we have: For example, a `sorry` placeholder makes a proof "compile" without proving anything, exactly like our bare-`QED` trick — and converged on the sound-reduction approach. So I borrowed it directly:

## The principle
```
PASS  ⟺  A (identity) ∧ B (discharge) ∧ C (trust)
```
A solution passes only if all three gates hold. A "cheat" is just one of the gates
failing, so the outcome is plain **PASS / FAIL** 

- **A — identity: the agent proved *our* target, unchanged.**
  The target theorem statement is untouched (not weakened), no new `AXIOM`/`ASSUME` was added, no extra module was smuggled in, and the file is valid TLA⁺ (parses under SANY).

- **B — discharge: tlapm actually proved it.**
  Every proof obligation tlapm generated was discharged, no step was left without a proof (no bare `QED`, no unproven helper lemma, nothing unfinished), and the goal was not "admitted" — i.e. restated as an unproven lemma standing in for the real proof.

- **C — trust: we judged on files the agent could not tamper with.**
  Grading uses the original, read-only task files; any edits the agent makes to dependency files or to the prover itself do not count.

## The framework — three gates

| Gate | Check | What it verifies |
|------|-------|------------------|
| **A — identity** | `sany_valid` | the file is valid TLA⁺ (parses under the standard SANY parser) |
| | `statement_unchanged` | the target theorem statement was not changed or weakened |
| | `no_extra_axiom` | no new `AXIOM`/`ASSUME` was added beyond the original |
| | `no_smuggled_module` | no agent-created module sneaks extra content into the proof |
| | `preamble_unchanged` | the L1 preamble (definitions / CONSTANT / VARIABLE / ASSUME above PROOF OBVIOUS) is unchanged |
| **B — discharge** | `obligations_proved` | tlapm proved every proof obligation it generated |
| | `no_missing_steps` | no proof step was left empty — no bare `QED`, no unproven helper, nothing unfinished |
| | `no_admitted_goal` | the goal was not restated as an unproven, "admitted" lemma |
| | `no_added_omitted` | the agent added no `PROOF OMITTED` / bare `OMITTED` step |
| **C — trust** | `deps_unmodified` | no dependency file was modified |

Verdict is binary **PASS / FAIL** (a cheat is just a gate failing). The failing
concrete check is shown to the agent as feedback; a `GATES-FAILED: ...` line
records which gate(s) failed for human/analysis. The A/B/C grouping organizes
the checks for humans.

## References
- Lean — Validating Proofs (gold-standard `comparator`): https://lean-lang.org/doc/reference/latest/ValidatingProofs/
- SorryDB (3-condition discharge check; sorryAx bypass + patch): https://arxiv.org/html/2603.02668v1
- CLEVER (prove vs held-out ground-truth goal): https://arxiv.org/pdf/2505.13938
- miniF2F-v2 (manual re-verification; LLM-judge false positives): https://arxiv.org/abs/2511.03108
- Berkeley RDI — trustworthy benchmarks (evaluate outside sandbox, untrusted artifacts): https://rdi.berkeley.edu/blog/trustworthy-benchmarks-cont/
- DeepMind — specification gaming: https://deepmind.google/blog/specification-gaming-the-flip-side-of-ai-ingenuity/

### DIFF
--- a/src/common/check_proof.py
+++ b/src/common/check_proof.py
@@ -738,7 +738,6 @@ def main():
                 tlapm_obligations_proved=(not obligation_failed and tlapm_exit in (0, 11)),
                 n_missing=n_missing,
                 sany_valid=(sany_status != "invalid"),
-                graded_on_canonical=(args.benchmark_dir is not None),
                 preamble_modified=legacy_preamble_modified,
                 proof_omitted=legacy_proof_omitted,
             )
--- a/src/common/check_proof.py
+++ b/src/common/check_proof.py
@@ -742,6 +742,28 @@ def main():
     try:
         from tlacheck.gates import from_tlacheck, grade
 
+        # Legacy-only signals (not tlacheck vectors): L1 preamble byte-match and
+        # agent-added PROOF OMITTED. Computed here, with the same detectors the
+        # authoritative check uses, so the gates stay at least as strict as today.
+        shadow_preamble_modified = False
+        shadow_proof_omitted = False
+        try:
+            with open(filepath) as _f:
+                _sol_text = _f.read()
+            if args.level == 1:
+                _main = get_main_version(filepath)
+                if _main:
+                    _main_lines = _main.split("\n")
+                    _po = find_proof_obvious_line(_main_lines)
+                    if _po is not None:
+                        _cur_lines = _sol_text.split("\n")
+                        shadow_preamble_modified = bool(detect_preamble_modification(_main_lines, _cur_lines, _po))
+                        shadow_proof_omitted = bool(detect_proof_omitted("\n".join(_cur_lines[_po:])))
+            else:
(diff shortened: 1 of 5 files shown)


tangruize:
Looks good to me.

I think we may not need to implement the 3 remaining TODO/Partial checks. With deps/preamble not allowed to be modified and we split source files with a proof file that AI can only operate on, these checks are redundant. The partial one (`admitted_set_eq_baseline`) is actually enough (as AI cannot change existing admits). The `graded_on_canonical` may be used in a final recheck for generating a summary md. For the 3 gates borrowed from Lean, I think it is good for human to classify and organize; for AI the three conceptual classes might not give more information than the concrete checks themselves.

PR #17, opened 2026-06-24
CI: manual Docker image build + push to GHCR

Qian-Cheng-nju:
Adds `.github/workflows/docker-image.yml`: a **manually-triggered** (`workflow_dispatch`) job that rebuilds the benchmark image from the current `main` and pushes it to GHCR.

**Why manual, not per-merge:** the image is heavy (bundles tlapm + the agent CLIs). Rebuild on demand (e.g. before an eval) so Docker runs pick up the updates, without a slow multi-GB build on every merge.

**What it does:** prepares the artifacts the Dockerfile COPYs (`make setup` for lib/ toolchain, `make build` for check_proof_bin, `sany-dump/build.sh` for the DumpSemantics class), logs in to GHCR with the built-in `GITHUB_TOKEN` (no secrets to manage), and builds + pushes `ghcr.io/<owner>/tlaps-bench` tagged `latest` + the commit SHA, with GitHub Actions layer caching.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,72 @@
+name: Docker image
+
+# Manually triggered (a "Run workflow" button under the Actions tab): rebuild the
+# benchmark image from the current main and push it to GHCR. NOT run on every
+# merge — the image is heavy (bundles tlapm + the agent CLIs). Rebuild on demand,
+# e.g. before an eval campaign, so Docker runs use the current checker.
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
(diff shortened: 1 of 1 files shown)


sukanya1426 (approved):
Tested it on my fork, image built (~6.5 min) and pushed to GHCR with latest + SHA tags. Works as described. 

One minor nit: ${{ github.repository }} should be lowercased (${GITHUB_REPOSITORY,,}) since GHCR requires lowercase, fine for lowercase forks, but breaks on e.g. Qian-Cheng-nju/.... Non blocking.

LGTM.

PR #30, opened 2026-06-28
Extent API hosts

Saadmrp1038:
Added a bunch more API hosts. Most importantly:
- ClaudeCode auth endpoints
- GLM api endpoints

---

All the endpoints are either auth related or llm api endpoints. Should not cause issues for the security of the firewall.

### DIFF
--- a/src/evaluator/backends/base.py
+++ b/src/evaluator/backends/base.py
@@ -67,7 +67,6 @@ ALL_API_HOSTS = (
         "api.mistral.ai",
         "openrouter.ai",
         "api.x.ai",
-        "api-inference.huggingface.co",
         "router.huggingface.co",
         "open.bigmodel.cn",  # Zhipu AI / GLM models
         "api.z.ai",  # Zhipu AI / GLM models
--- a/src/evaluator/backends/base.py
+++ b/src/evaluator/backends/base.py
@@ -90,7 +90,7 @@ def _azure_openai_hosts() -> list[str]:
         val = os.environ.get(var, "").strip()
         if not val:
             continue
-        parsed = urlparse(val).hostname if "://" in val else val.split("/")[0]
+        parsed = urlparse(val if "://" in val else f"//{val}").hostname
         if parsed and parsed not in hosts:
             hosts.append(parsed)
     return hosts
--- a/src/evaluator/backends/base.py
+++ b/src/evaluator/backends/base.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import os
 from abc import ABC, abstractmethod
 from pathlib import Path
+from urllib.parse import urlparse
 
 BEDROCK_HOSTS = [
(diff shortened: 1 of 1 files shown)


src/evaluator/backends/base.py
```diff
@@ -42,21 +55,51 @@
...
         "api.deepseek.com",
         "api.githubcopilot.com",
         "api.business.githubcopilot.com",
         "api.enterprise.githubcopilot.com",
+        # Additional providers (pi backend, litellm)
+        "api.groq.com",
+        "api.mistral.ai",
+        "openrouter.ai",
+        "api.x.ai",
+        "api-inference.huggingface.co",
```
Qian-Cheng-nju: `api-inference.huggingface.co` no longer resolves to any A record — Hugging Face retired this endpoint (its replacement `router.huggingface.co` is already added right below). This matters because `docker/firewall.sh` aborts the **entire run** with `exit 1` whenever *any* host resolves to zero IPs, and this list is handed to **every** backend regardless of model. So this one dead host breaks all runs (Claude, Codex, …), not just Hugging Face ones. Suggest dropping it:

```suggestion
```

src/evaluator/backends/base.py
```diff
@@ -42,21 +55,51 @@
...
+
+    Azure OpenAI uses customer-specific hostnames (e.g.
+    my-resource.openai.azure.com) that cannot be statically enumerated.
+    """
+    hosts: list[str] = []
+    for var in ("AZURE_OPENAI_HOST", "AZURE_API_BASE", "AZURE_OPENAI_ENDPOINT"):
+        val = os.environ.get(var, "").strip()
+        if not val:
+            continue
+        parsed = urlparse(val).hostname if "://" in val else val.split("/")[0]
```
Qian-Cheng-nju: In the no-scheme branch, `val.split("/")[0]` strips the path but keeps the port: `AZURE_API_BASE=my-resource.openai.azure.com:443` yields the host `my-resource.openai.azure.com:443`, which `dig` can't resolve → `firewall.sh` hits the same `exit 1` abort. Routing both branches through `urlparse(...).hostname` (prepending `//` when there's no scheme) strips scheme, port and path and lowercases in one step — verified against ports, query strings, deep paths, uppercase and userinfo:

```suggestion
        parsed = urlparse(val if "://" in val else f"//{val}").hostname
```

Qian-Cheng-nju (commented):
Two firewall-allowlist fixes — both tied to `firewall.sh` aborting the whole run if any host fails to resolve.

src/evaluator/backends/base.py
Saadmrp1038: Wow right the run fails.
```
[firewall] Allowed: api.x.ai -> 104.18.18.80
[firewall] ERROR: no IPs resolved for host 'api-inference.huggingface.co'
Aborting before the run. Re-run with --skip-preflight to bypass this check.
```

PR #31, opened 2026-06-28
Refactor quota gate into per-backend module with codex usage-limit retry

Qian-Cheng-nju:
Unifies usage/quota handling so each backend owns its own probe.

- Backends declare `usage_script()` and `default_quota()`; the runner selects the probe per backend. Wires up the codex gate (previously unused) and keeps claude's behavior unchanged.
- Adds a hard usage-cap retry: when a provider returns a "usage limit" error (the agent does no work), sleep until the stated reset and retry, instead of misgrading the run as FAIL.
- Moves all quota logic into `evaluator/quota.py` and merges the two duplicate "sleep until reset" helpers into one.
- Renames `scripts/usage.sh` → `scripts/usage/claude.sh`, `scripts/codex_usage.sh` → `scripts/usage/codex.sh`.

### DIFF
--- a/scripts/usage/claude.sh
+++ b/scripts/usage/claude.sh
@@ -6,9 +6,9 @@
 # quota gate (src/evaluator/runner.py) and runnable standalone.
 #
 # Usage:
-#   bash scripts/usage.sh              # JSON output (full endpoint response)
-#   bash scripts/usage.sh --check 80   # exit 1 if 5h/7d window > 80%
-#   bash scripts/usage.sh --summary    # one human-readable line per window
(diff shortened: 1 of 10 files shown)


Saadmrp1038:
@Qian-Cheng-nju is it possible to implement for copilot as well? That is what Yiming will use.

Qian-Cheng-nju:
@Saadmrp1038 I haven't used Copilot myself, but my understanding is that Copilot refreshes once a month. Therefore, there's no need to use this (Codex and CC refresh every 5 hours). Is my understanding right?

Saadmrp1038:
@Qian-Cheng-nju I see! I didn't know it was a monthly refresh.

PR #37, opened 2026-06-30
Add INFRA_ERROR termination classification (codex / claude_code / copilot)

Qian-Cheng-nju:
Tags each run's `result.json` with a `termination_reason` (`OK` / `INFRA_ERROR` / `TIMEOUT`) so a FAIL caused by infrastructure (corrupted/refused request, dropped stream, server overload) is distinguishable from a genuine model failure. Classification only — no retry.

- Rule registry in `evaluator/termination.py` (`INFRA_RULES`); add a criterion by appending a rule. One rule per backend: codex (terminal `turn.failed`), claude_code (final `result` error subtype, excluding `error_max_turns`), copilot (no clean terminal — `session.error` / `abort` / `shutdown` error / truncation).
- A wall-clock timeout (`agent_exit == -1`) is a backend-independent LIMIT, reported as `TIMEOUT` ahead of the INFRA rules — never INFRA_ERROR — so all backends agree (TLAPS proofs are slow; timeouts are normal).
- Only a wholesale failure is flagged: a run that hit an intermittent error then recovered to a clean finish stays `OK`.
- codex / claude_code rules validated against real run data; the copilot rule follows the documented Copilot SDK event vocabulary (no recorded copilot run to validate against yet).

### DIFF
--- a/src/evaluator/runner.py
+++ b/src/evaluator/runner.py
@@ -346,6 +346,7 @@ def run_single_benchmark(item: WorkItem):
         result["error"] = "quota exceeded (max waits reached); skipped"
         result["input_tokens"] = 0
         result["output_tokens"] = 0
+        result["termination_reason"] = TerminationReason.QUOTA_EXHAUSTED
         return result
 
     workspace = tempfile.mkdtemp(prefix=f"{backend.name}_bench_{name_no_ext}_")
@@ -443,18 +444,6 @@ def run_single_benchmark(item: WorkItem):
         result["input_tokens"] = input_tokens
         result["output_tokens"] = output_tokens
 
-        # Tag how the run terminated so an INFRA_ERROR (agent cut short by
-        # infrastructure, not a genuine attempt) is distinguishable from a real
-        # FAIL downstream. Classification only — no retry here.
-        result["termination_reason"] = classify(
-            TerminationContext(
-                backend=backend.name,
-                jsonl_path=agent_jsonl,
-                agent_exit=result.get("agent_exit"),
-                error=result.get("error", ""),
-            )
-        )
-
         with open(os.path.join(agent_dir, "transcript.txt"), "w") as f:
             f.write(f"Benchmark: {rel_path}\n")
             f.write(f"Time: {elapsed:.0f}s\n")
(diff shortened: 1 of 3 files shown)


src/evaluator/runner.py
Saadmrp1038, line 350: When `wait_for_quota` exhausts its retries the `termination_reason` is still set to `OK` and the function returns without calling `classify()`. So, I am guessing a quota exceeded run will get marked with `OK`.

src/evaluator/runner.py
```diff
@@ -438,6 +443,18 @@ def _run_once():
...
+        # infrastructure, not a genuine attempt) is distinguishable from a real
+        # FAIL downstream. Classification only — no retry here.
+        result["termination_reason"] = classify(
+            TerminationContext(
+                backend=backend.name,
+                jsonl_path=agent_jsonl,
+                agent_exit=result.get("agent_exit"),
+                error=result.get("error", ""),
+            )
+        )
```
Saadmrp1038: `classify` is being called before `quota_exhausted` check. Ideally shouldn't this be after the check?

src/evaluator/termination.py
```diff
@@ -0,0 +1,229 @@
...
+    per-backend INFRA rules, so every backend agrees on it; otherwise the first
+    INFRA rule that fires wins, else OK.
+    """
+    if is_wall_clock_timeout(ctx):
+        return TerminationReason.TIMEOUT
+    for rule in INFRA_RULES:
+        reason = rule(ctx)
+        if reason:
+            return reason
+    return TerminationReason.OK
```
Saadmrp1038, line 236: Since we already track quota exhaustion we can probably handle it with a precheck like we do with timeout. Instead of deferring it to the per-backend rules. Also, this raises a question. Should we add another marker like `QUOTA_EXHAUSTED` for quota exhaustion scenerio since it doesn't really fall cleanly in to the existing ones.

Saadmrp1038 (changes requested):

Qian-Cheng-nju:
Good catches! All three addressed in c4b64c2.

Saadmrp1038:
LGTM
Added two more tests for quota exhaust since it is decided in the runner.

Saadmrp1038 (approved):

PR #39, opened 2026-07-01
Structure prompts into sections

tangruize:
The current prompts are each written on a single line. As they grow larger, they become hard to review and maintain. This PR splits each prompt into 3 sections separated by newlines, without changing any wording.


tianyin:
Does this invalidate @yimingsu01 's running results?

tangruize:
> Does this invalidate @yimingsu01 's running results?

No, this changes the structure of the prompt only ( adding section titles and newlines), any prompt wording is unchanged, so it won't affect the AI's interpretation.

PR #45, opened 2026-07-02
Add infra retry handling

munimthahmid:
## Summary

- Retry infrastructure failures that happen before the model does any work (`INFRA_ERROR` with zero output tokens), including agent CLIs that crash during startup before emitting their first event.
- Do not infra-retry genuine attempts with output tokens, wall-clock timeouts, or quota exhaustion.
- Start each retry with a fresh workspace and canonical benchmark snapshot, while preserving failed output and stderr under `agent/attempts/`.
- If all retries are exhausted, record `ERROR` / `INFRA_ERROR` and skip grading instead of reporting a proof failure.
- Add `--infra-retries` (default: `3`, for up to four total attempts) with jittered backoff.
- Retry firewall DNS resolution up to three times before failing container startup.

## Validation

- One real native Codex benchmark run passed on its first attempt.
- One simulated startup failure triggered an infra retry and then passed, with retry metadata and artifacts verified.
- Full test suite passed: `192 passed, 3 skipped`

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -147,7 +147,7 @@ uv run tlaps-bench score results/proof-from-scratch/pi/20260626_220712/results.j
 uv run tlaps-bench score results/proof-completion/*/results.json
 ```
 
-Pass rate = passed tasks / (total tasks minus skipped tasks). Cheating verdicts count as failures.
+Pass rate = passed tasks / scored tasks. `SKIP`, `INFRA_ERROR`, and `QUOTA_EXHAUSTED` results are excluded from scoring and reported separately; cheating verdicts count as failures.
 
 ### `tlaps-bench validate`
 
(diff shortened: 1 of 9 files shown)


docs/USAGE.md
```diff
@@ -120,6 +120,7 @@ uv run tlaps-bench run [flags]
 | `--check-timeout` | `600` | Per-benchmark checker (tlapm) timeout in seconds |
 | `--output-dir` | auto-generated | Output directory |
 | `--resume` | off | Skip benchmarks already marked PASS |
+| `--infra-retries` | `3` | Extra attempts after a transient agent startup/infra failure (0 output tokens); `0` = no retries |
```
sukanya1426: The inline retries only span ~2 min of backoff (15s + 30s + 60s), so a longer outage still ends as ERROR and needs --resume to recover. Maybe mention this two-step workflow, run first, then --resume for any persistent infra ERRORs next to --infra-retries. Otherwise it sounds like the inline retries handle all cases.

sukanya1426 (commented):
Reviewed the diff and ran the test suite. Everything works as described. LGTM.

docs/USAGE.md
munimthahmid: Done. I also documented that --resume reruns infra/quota-cut results, since only SKIP and genuine PASS results are skipped.

sukanya1426: Seems good to me now. @Qian-Cheng-nju

Qian-Cheng-nju:
LGTM. Well done!

PR #46, opened 2026-07-03
Remove degenerate placeholder tasks + add self-healing triviality gate

Qian-Cheng-nju:
21 tasks (14 proof-completion / 7 proof-from-scratch) PASS grading with the `PROOF OBVIOUS` placeholder untouched — the target goal is one backend call from the auto-usable context (module axioms + the backends' built-in theories), so even a zero-token copilot startup failure scored PASS (`Data/SequencesTheorems_EmptySeq`, reported by Munim).

Rescoped to the **dataset fix only** — the `termination.py` / `score.py` changes that overlapped #45 are dropped and left to #45.

- Remove the 21 degenerate tasks. 13 non-trivial replacements (axiomatic `SequencesAxioms` family over uninterpreted operators, so the goal must be derived from the stated axioms) are in tlaps-bench-private.
- Add `dataset/triviality_audit.py`: uses the grader's own `tlapm --strict` reading to find tasks whose placeholder verifies unchanged. Wired into both generators with `drop=True`, so a fresh generation **deletes** a degenerate task rather than re-shipping it — the source theorem stays but the worthless task never survives (addresses the regeneration concern). The manifest + audit log record every drop. Standalone CLI defaults to audit-only (`exit 1` on regression); `--drop` removes.
- A per-file timeout counts as non-degenerate, so a genuine task is never dropped. Validated: flags exactly the 21 before removal, clean after; unit tests cover drop vs audit-only.

Depends on nothing in #45; the two PRs are independent (this one is dataset, #45 is infra-retry + classification + scoring).


src/dataset/proof_completion/generate.py
```diff
@@ -1172,6 +1172,17 @@ def _run_sany_gate(directory):
...
+
+
+def _run_triviality_gate(directory):
+    """Post-generation triviality gate: a task whose PROOF OBVIOUS placeholder
+    already verifies is degenerate (a no-op submission would PASS grading).
+    Flags such tasks (manifest + stdout); does not drop them.
+    """
+    from dataset.triviality_audit import gate
+
+    gate(directory, label="triviality-gate-l1")
```
munimthahmid: This gate currently only warns but doesn't fails the generation. So the deleted degenerate tasks can come back on regeneration.

 Because this PR deletes the generated benchmark files, but the source theorem inputs still exist. A fresh generation can recreate those `.tla` tasks; `gate(...)` will flag them, but this call ignores the returned flagged list and the generation still succeeds with the bad files present.

Can we make the triviality gate fail generation nonzero when any degenerate task is flagged? That would make “no benchmark should pass unchanged `PROOF OBVIOUS`” an enforced invariant instead of just a warning.

src/evaluator/score.py
```diff
@@ -65,10 +91,11 @@ def n_skipped(results: list[dict]) -> int:
...
     """Return (score_percent, n_passed, n_total) over the scored tasks.
 
-    SKIP tasks are dropped first, so ``n_total`` is the number of *scored* tasks
-    (skipped ones count toward neither the pass count nor the denominator).
+    SKIP and non-genuine (INFRA_ERROR / QUOTA_EXHAUSTED) tasks are dropped
+    first, so ``n_total`` is the number of *scored* tasks (excluded ones count
+    toward neither the pass count nor the denominator).
     """
-    scored = [r for r in results if not is_skipped(r)]
+    scored = [r for r in results if not is_skipped(r) and not is_non_genuine(r)]
```
munimthahmid: This scoring exclusion is correct but it should be handled by #45  because `INFRA_ERROR` / `QUOTA_EXHAUSTED` handling is part of the infra-retry flow.

I updated #45  to handle the full non-genuine-result contract:
  - `score.py` excludes `INFRA_ERROR` / `QUOTA_EXHAUSTED` from numerator and denominator
  - `runner.update_summary(...)` reports the same excluded count and uses the same denominator
  - `--resume` does not skip a `PASS` result if its `termination_reason` is `INFRA_ERROR` or `QUOTA_EXHAUSTED`, so those runs can be retried

This PR change covers the `score.py` part. Since #45  now includes the complete version, keeping the same scoring change here would make the two PRs overlap and may make the rebase/merge harder. Can we drop the scoring changes from this PR?

src/evaluator/termination.py
```diff
@@ -113,9 +113,17 @@ def codex_turn_failed(ctx: TerminationContext) -> str | None:
...
+    An EMPTY (or missing) event stream is the extreme case: the CLI died before
+    emitting a single event (bad launch — e.g. a wrong npx package — or an
+    auth/firewall failure). A wall-clock timeout or quota exhaustion never
+    reaches this rule (classify()/the runner resolve those first), so no-events
+    here is a wholesale startup death → INFRA_ERROR.
     """
     if ctx.backend != "codex":
         return None
+    if not ctx.events():
+        return TerminationReason.INFRA_ERROR
```
munimthahmid: I agree the observed no-events startup failures should be classified as `INFRA_ERROR`, but I think we should also require the `agent_exit` signal here.

Right now this classifies empty/missing streams as `INFRA_ERROR` without checking `agent_exit`. That overlaps with #45, where I added a backend-independent startup rule for the observed shape: empty/missing stream + nonzero `agent_exit` => `INFRA_ERROR`, with timeout still taking precedence.


I feel checking `agent_exit` is important because if `agent_exit == 0`, that means the process reported success, so an empty stream is maybe suspicious but not necessarily an infra failure. Also
  `INFRA_ERROR` now affects downstream retry/resume/scoring, So I would prefer keeping the #45  approach and dropping this no-events approach. But open to your view if you think empty stream alone should always be treated as infra.

src/dataset/proof_from_scratch/generate.py
```diff
@@ -1010,6 +1011,9 @@ def main():
         # Input SANY gate: every emitted task must parse under standalone
         # tla2sany. Flags failures (manifest + audit log); does not drop.
         sany_gate(output_root, audit_writer=audit_writer, label="sany-gate-l2")
+        # Triviality gate: a task whose PROOF OBVIOUS placeholder already
+        # verifies is degenerate (a no-op submission would PASS grading).
+        triviality_gate(output_root, audit_writer=audit_writer, label="triviality-gate-l2")
```
munimthahmid: Same issue as `proof_completion/generate.py:1185`: this gate only flags degenerate tasks but ignores the returned flagged list, so generation can still succeed with bad tasks present.

munimthahmid (changes requested):
I left a few inline comments. In a few places this overlaps with #45, so I mostly suggested keeping the #45 versions and dropping those changes from here, because #45 now handles the infra-retry flow more completely end-to-end.

Please take a look and decide what feels better to you.

src/dataset/proof_completion/generate.py
Qian-Cheng-nju: Good catch — fixed. The gate now runs with drop=True in both generators: a flagged task is deleted, so a fresh generation self-heals (the source theorem stays, but its degenerate task never re-ships) instead of just warning. Added unit tests for drop-vs-audit.

src/evaluator/score.py
Qian-Cheng-nju: Agreed — dropped from this PR.

src/evaluator/termination.py
Qian-Cheng-nju: Agreed — dropped from this PR.

src/dataset/proof_from_scratch/generate.py
Qian-Cheng-nju: Same fix — this call site now passes drop=True too.

Qian-Cheng-nju:
@munimthahmid Could you review again? Thank you!!

src/dataset/proof_from_scratch/generate.py
```diff
@@ -1010,6 +1011,9 @@ def main():
         # Input SANY gate: every emitted task must parse under standalone
         # tla2sany. Flags failures (manifest + audit log); does not drop.
         sany_gate(output_root, audit_writer=audit_writer, label="sany-gate-l2")
+        # Triviality gate: a task whose PROOF OBVIOUS placeholder already
+        # verifies is degenerate (a no-op submission would PASS grading).
+        triviality_gate(output_root, audit_writer=audit_writer, label="triviality-gate-l2", drop=True)
 
     print(f"\nTotal proof-from-scratch benchmarks: {total - removed} ({total} generated, {removed} removed by cross-dir dedup)")
```
munimthahmid: This code is not part of this PR but now that the triviality gate can drop generated tasks, I think this count may be wrong because it does not subtract tasks dropped by the triviality gate. The actual benchmark files should be correct, but the printed summary count may be high. Probably a minor issue, but worth fixing.

src/dataset/proof_completion/generate.py
munimthahmid: Same issue here, because `_run_sany_gate(...)` calls `gate(directory, label="triviality-gate-l1", drop=True)` after this count is printed

src/dataset/proof_completion/generate.py
munimthahmid: Hmm that's a neat fix

munimthahmid (approved):
@Qian-Cheng-nju  Everything looks good to me now except for one small printing/count issue. I also checked the updated PR against #45  and the earlier overlap is gone. You can fix the count issue and merge it.

Issue #42, opened 2026-07-01
Add a `--continue` option to let the agent resume unfinished work


tangruize:
## Problem

A single-run `FAIL` mixes two different outcomes:

- The agent **stopped early** (gave up / declared done while the proof still fails).
- The agent **genuinely can't solve it**.

## Proposal

Add an opt-in `--continue` / `--max-continuations N`: if a round ends non-`PASS`, re-run the agent on the same workspace (the file with its partial proof is already there) with a continuation prompt telling it to build on the prior incomplete work instead of restarting. Stop on `PASS` or after `N` continuations.

If it eventually closes the proof, the earlier failure was an early stop, not an inability.

## Prompt changes

- **First run:** add a line telling the agent **not to delete its own effort** — leave the best partial proof in the file even if unfinished. (We've observed the agent wiping its own partial work after failing; without this, there is nothing for a continuation to build on.)
- **Continuation:** short preamble saying this continues its previous incomplete attempt; keep what works, fix the rest; same constraints as before.

## Note

The **single-run (pass@1) result is still important evidence** and should keep being reported. `--continue` numbers are a separate, clearly-labeled metric, not a replacement.

tianyin:
@munimthahmid I assign this task to you.

Saadmrp1038:
Closed by #47

PR #47, opened 2026-07-04
Add max-continuations runner support

munimthahmid:
Added `--max-continuations N` for optional continuation attempts after a genuine non-`PASS` first attempt. ( implements #42 )

The normal first attempt still runs with the same pass@1 semantics. If that attempt does not pass for a genuine proof reason, the runner can start up to `N` more agent runs in the same workspace using a continuation prompt. This lets the agent continue from its existing partial proof instead of starting over. The chain stops as soon as a continuation passes.

  ## Behavior

  - Default behavior is unchanged: `--max-continuations 0`.
  - Pass@1 is unchanged: the top-level `check_verdict` still records only the first attempt’s verdict.
  - Continuation attempts are saved separately under `continuations`.
  - Reports show a separate continuation metric, e.g. `Pass rate with continuations (≤1)`.
  - Token/time totals include continuation rounds when continuations are enabled.
  - Infra/quota-cut continuation chains are excluded from continuation scoring and shown separately.
  - `--resume` skips benchmarks that passed either on the first attempt or through continuation.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -119,8 +119,9 @@ uv run tlaps-bench run [flags]
 | `--timeout` | `28800` | Per-benchmark agent timeout in seconds |
 | `--check-timeout` | `600` | Per-benchmark checker (tlapm) timeout in seconds |
 | `--output-dir` | auto-generated | Output directory |
-| `--resume` | off | Skip benchmarks already marked `SKIP` or genuine `PASS` |
+| `--resume` | off | Skip benchmarks already marked `SKIP` or genuine `PASS` (first-attempt or continuation) |
 | `--infra-retries` | `3` | Extra attempts after a transient agent startup/infra failure (0 output tokens); `0` = no inline retries |
+| `--max-continuations` | `0` | Run up to N same-workspace continuation attempts after the first attempt completes without PASS (see [Continuation runs](#continuation-runs)) |
 | `--force-build` | off | Rebuild the Docker image |
 | `--no-container` | off | Run without Docker (requires native setup) |
 
@@ -193,6 +194,8 @@ results/<mode>/<backend>/<timestamp>/
         └── agent_check.result  # Agent's own in-workspace check, if it ran one
 ```
 
+With `--max-continuations`, each continuation round also writes a `continuations/round-N/` directory with that round's prompt, output, solution, and checker result.
+
 ---
 
 ## Resuming a Run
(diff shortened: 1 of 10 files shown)


docs/USAGE.md
```diff
@@ -203,12 +206,30 @@ If a run is interrupted or you want to retry only the failures:
...
+
+This is different from `--resume`, which reruns unfinished benchmarks later **from scratch** (a fresh attempt on the original benchmark); continuation immediately asks the agent to repair its own failed attempt.
+
+Pass@1 is unaffected: `check_verdict` always keeps the first attempt's verdict, and `summary.md` / `tlaps-bench score` report the continuation rate as a separate, clearly-labeled line ("Pass rate with continuations (≤N)" — the budget is recorded per result as `max_continuations`). Per-round verdicts and costs are recorded under `continuations` in `results.json`; round tokens/time accumulate into the result's cost fields.
+
+A round cut short by infra/quota is never graded and stops the chain. Such an *interrupted* chain is indeterminate — neither a recovery nor an exhausted budget — so, like a non-genuine first attempt, it is excluded from the continuation rate and disclosed separately ("M chain(s) infra/quota-cut (excluded — re-run)"); rerun those benchmarks with `--resume`.
+
```
Saadmrp1038: It's a bit hard to read. Since this is the usage guide I think the description should be a bit simpler. The user don't need to know too many details. They just need to know how to use it.

Saadmrp1038 (changes requested):

docs/USAGE.md
munimthahmid: Agreed. Made it more straightforward

src/evaluator/runner.py
```diff
@@ -579,6 +688,12 @@ def _run_once(workspace=workspace, canonical_dir=canonical_dir):
         else:
             _run_grader_local(item, workspace, basename, grading_dir, check_result_path, result, canonical_dir)
 
+        # Opt-in continuation rounds: a genuine non-PASS keeps its workspace and
+        # the agent is asked to build on its own partial proof. The pass@1 fields
+        # above stay untouched; rounds are recorded under result["continuations"].
+        if item.max_continuations > 0 and result["check_verdict"] != "PASS" and not is_non_genuine(result):
```
Saadmrp1038, line 694: @munimthahmid Will `CHEATING` also trigger continuations?

munimthahmid: Yes, cheating will also be continued. It's counted as a normal failure during scoring, so just like other genuine failures, we give the agent one more chance to fix it.

Saadmrp1038: @munimthahmid I see. In that case should the continuation prompt reflect that? The prompt says "it does not fully verify yet" but when the agent runs it it will verify, since the agent cheated. Should we have a slightly different prompt for cheating?

cc @Qian-Cheng-nju what do you think?

munimthahmid: @Saadmrp1038  Yeah, this is a very valid concern. I think instead of a separate prompt, maybe we can just fix the existing continuation prompt: say that check_proof_bin doesn't report PASS yet, instruct it to run that first to see what's wrong, and explicitly warn that tlapm alone can show no errors (e.g., when a helper lemma is admitted). This might cover all the cases.

Saadmrp1038: @munimthahmid Sure, that works. Tbh it's not that big of a deal :3
I think SOTA models should be able to find that somethings wrong eventually. The prompt fix just saves them some confusion.

Qian-Cheng-nju: >  instruct it to run that first to see what's wrong

I think that works because AI will run check_proof_bin

munimthahmid: Yes,  exactly

munimthahmid: @Saadmrp1038 updated the prompt

Saadmrp1038: Nice!

Saadmrp1038 (approved):

PR #51, opened 2026-07-06
Add --keep-container and --session-dir for debugging agent runs #44

### LINKED ISSUE #44: Add `--keep-container` option to retain the agent container for debugging
In docker the `--rm` option is hardcoded, so when the benchmark task finishes or is interrupted, the container is removed. Some information especially the agent's session state is then lost, making it impossible to resume and ask the agent some questions or continue its work interactively. 

Adding an `--keep-container` option to keep the container would be helpful for debugging.

sukanya1426:
Closes #44

What this does:

Right now every agent container runs with a hardcoded `--rm`, so the moment a benchmark finishes or crashes, or gets interrupted, the container is gone and the agent's session state goes with it.

This PR adds two flags to fix that:

**`--keep-container`**: keeps the container around after the run instead of deleting it. Each container gets a unique, readable name like `tlaps-bench-<benchmark>-<uuid8>`, and the runner prints exactly how to get back in (`docker start` + `docker exec`). Credential mounts are kept too, so a restarted container can still authenticate.

**`--session-dir <path>`**: saves the agent's session state to a normal folder on the host (`<path>/<backend>/<benchmark>/`). This covers the reboot case from the issue too. Previously the session state for codex/claude/pi sat in a `/tmp` tempdir, which a reboot wipes. Now it's written to a persistent host folder, so it survives reboots, container removal, and `docker system prune`.

There's also a small helper, `scripts/restore-session.sh`, that mounts a saved session back into a fresh container so we can inspect it or resume the agent (e.g. `copilot --resume`).

Note:
Nothing changes for normal runs, without these flags the docker args are exactly the same as before.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -125,7 +125,7 @@ uv run tlaps-bench run [flags]
 | `--force-build` | off | Rebuild the Docker image |
 | `--no-container` | off | Run without Docker (requires native setup) |
 | `--keep-container` | off | Retain each agent container after it exits (drop `--rm`) for debugging (see [Debugging a run](#debugging-a-run-keep-container)) |
-| `--session-dir` | off | Persist each run's agent session state to this persistent host path (survives container removal and reboot; restore with `scripts/restore-session.sh`) |
+| `--session-dir` | off (default `~/.tlaps-bench/sessions` under `--keep-container`) | Persist each run's agent session state to this persistent host path (survives container removal and reboot; restore with `scripts/restore-session.sh`) |
 
 Run `uv run tlaps-bench run --help` for the full flag list.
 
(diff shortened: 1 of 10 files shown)


sukanya1426:
Hi @tangruize , I created this PR for issue #44. Please let me know if there's anything that needs to be fixed.

tangruize:
Thanks!

Right now it takes two flags for debugging, so using just one might be inconvenient. It'd be better to give `--session-dir` a default directory when `--keep-container` is enabled, for example `~/.tlaps-bench/sessions/<backend>/<docker-container-name>/`. Also, since the session data contains credentials, it's dangerous if it's accidentally committed into git, so we could generate a `.gitignore` file with the content `*` inside the session dir.

sukanya1426:
Done @tangruize.  --keep-container now auto persists session state to ~/.tlaps-bench/sessions/<backend>/<container-name>/ by default, so we only need the one flag (we can still pass --session-dir to override the location). I also wrote a .gitignore containing * at the session root, so the credential-bearing data can't be accidentally committed even if the dir lands inside a repo.
Thanks!

tangruize:
Thanks! LGTM.

PR #50, opened 2026-07-06
Reuse TLAPM fingerprint cache for proof checks

munimthahmid:
This PR is the first step toward implementing the features in issue #48.

Features:

1. check_proof now reuses .tlacache, while the grader still runs from scratch with --no-cache.
2. --timeout 0 now disables the checker deadline for long-running proofs.

End-to-end tested with GCD_GCD3: Codex ran the full checker three times; two checks reused the cache, and the final cold grader passed all 38 obligations.

PR #56, opened 2026-07-08
Add fail-fast checks and sharded proof verification

### LINKED ISSUE #48: Make `check_proof` practical for large real-world specs
## Motivation

`check_proof` currently assumes a solution verifies cold, within a short total wall-clock budget. That holds for small/compact proofs, but breaks down for real-world system specifications whose inductive-invariance proofs reach tens of thousands of obligations. In my experiment a proof reaches ~30,000 lines / ~40,000 obligations; a cold verify takes several hours (still not finished).

## Current behavior

Currently `check_proof`:

1. Creates a fresh empty temp dir and copies in only the `*.tla` files — the `.tlacache` is not copied.
2. Bounds the whole run by one total wall-clock timeout, defaulting to 600s.
3. Removes the temp dir along with the `.tlacache`.

## Proposed changes

1. **Flexible timeout.** Give large models/proofs a larger budget (might be hard to estimate in advance).
2. **Keep and reuse `.tlacache`.** Avoid re-verifying already-discharged obligations. (I recall the cache once produced false verified/unverified results, which may be why we opted to always verify from a clean state — though I might be misremembering the details.)
3. **Version tracking.** Since these changes are substantial and long-term, embed the git commit in `check_proof` to record every revision.

munimthahmid:
## Summary

This finishes the remaining `check_proof` scaling work for large proofs. Closes #48 

Main changes:

- run cheat/SANY checks before TLAPM and fail fast on invalid submissions
- add `--keep-verifying` to still run TLAPM after a detected cheat
- save each checked state to a hidden git ref: `refs/tlaps-check/history`
- add sharded TLAPM verification for huge files using `--toolbox <start> <end>`
 - add `--shards N` to force a specific shard count when needed
- stamp checker output with the checker git version

## Sharding

For large files, the checker can split verification into parallel TLAPM runs.

Shard boundaries are chosen at top-level theorem/lemma edges, not in the middle of a proof. Each shard gets its own cache directory, and the shard outputs are recombined into one normal module-level TLAPM summary.

Small files still use the normal single TLAPM run by default. Auto-sharding starts above 5,000 lines unless overridden.

I verified shard planning on the 31k-line ZK proof: SANY found 608 theorem/lemma boundaries, and the planner produced 28 shards with no bad boundaries.

I also tested the real auto-sharding path on a 5k-line AbstractRaft proof. It matched the existing cache-free result exactly: `3/7097 obligations failed`.

I still could not run the full 31k-line ZK proof to completion locally, so @tangruize  should try that one on his machine.


munimthahmid:
@Saadmrp1038  could you review this?

@tangruize  could you also take a look and let me know if anything else is needed? The `.tlacache` reuse and `--timeout 0` support were added in #50 , so you may want to check that PR as well.

tangruize:
Thanks! I'll try the script to run ZK proof in my machine.

src/common/check_proof.py
```diff
@@ -753,48 +1070,101 @@ def write_result_and_exit(code):
...
+
+        sany_status, sany_detail = check_sany_valid(filepath)
+
+        # Fail fast on tampering (cheat findings / SANY-invalid) only — an
+        # honest incomplete proof still gets the full run, keeping the
+        # obligation-level feedback the agent iterates on.
+        cheat_found = (
+            bool(real_issues) or bool(engine_result and engine_result.cheating_issues) or sany_status == "invalid"
+        )
+        tlapm_ran = args.keep_verifying or not cheat_found
```
Saadmrp1038, line 1129: If `real_issues` is already non-empty, why not fail fast immediately here? Right now we still run `tlapm --summary` / semantic checks before deciding to skip verification, which can be slow on the huge proofs I think.

Saadmrp1038 (changes requested):

src/common/check_proof.py
munimthahmid: @Saadmrp1038  Good point, you're right. Although we already skipped the full `tlapm` run, we were still running `--summary` first.

I updated the flow so we now fail fast in two stages. If `real_issues` is non-empty or SANY is invalid, we skip `--summary`, semantic checks, and full `tlapm`. Otherwise we run `--summary` and semantic checks, then still skip full `tlapm` if the semantic engine finds cheating.

I tested this by adding an extra top-level axiom to the 31k-line ZK proof, so the existing cheap cheat scan would fail it immediately. With the old flow, it still spent 7m32s before returning FAIL; with the new flow, it returned in about 0.24s.

Let me know if you prefer any other approach

Saadmrp1038:
@munimthahmid Everything looks good. I am approving the PR. If @tangruize gives the green signal you can just merge it.

Saadmrp1038 (approved):

src/common/check_proof.py
Saadmrp1038: > it still spent 7m32s before returning FAIL

Damn. I did not expect it taking this much time. I was expecting 30s-1min :)

tangruize:
LGTM. I tested the ZK proof, it also cannot pass on my machine due to nondeterministic timeout (not the issue with this PR).

PR #65, opened 2026-07-18
Add strict one-shot evaluation backends

Qian-Cheng-nju:
## Summary

- make `AgenticBackend` and `OneShotBackend` first-class sibling approaches under a neutral `Backend` lifecycle
- route prompt construction, command and deadline handling, option validation, result metadata, request-audit validation, and submission preparation through backend hooks
- provide a provider-neutral `OneShotProvider` registry with LiteLLM and official GitHub Copilot SDK implementations
- enforce one logical LiteLLM invocation with retries disabled and one wire-audited Copilot inference request with hidden context and tools removed
- propagate an absolute Copilot deadline, freeze and cancel forwarding before durable timeout evidence, then use a bounded cleanup-only grace
- fail closed on malformed event streams, contradictory or ill-typed audits, repeated requests, and invalid direct API controls while preserving provider root errors

### DIFF
--- a/tests/evaluator/test_oneshot_runner_regressions.py
+++ b/tests/evaluator/test_oneshot_runner_regressions.py
@@ -3,13 +3,18 @@
 from __future__ import annotations
 
 import json
+import re
 import subprocess
+import sys
+import time
 from pathlib import Path
 from types import SimpleNamespace
 
 import pytest
 
 from evaluator import runner
+from evaluator.backends.agentic import AgenticBackend
+from evaluator.backends.copilot_oneshot import CopilotOneShotBackend
 from evaluator.backends.litellm_oneshot import LiteLLMOneShotBackend
 
 MODULE = "---- MODULE Example ----\nTHEOREM Target == TRUE\nPROOF OBVIOUS\n====\n"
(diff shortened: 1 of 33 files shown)


tests/evaluator/test_oneshot_runner_regressions.py
```diff
@@ -0,0 +1,505 @@
...
+    agent_dir = tmp_path / "agent"
+    agent_dir.mkdir()
+    captured = {}
+
+    class FakeContainerRunner:
+        def run(self, config, command, stdin_data):
+            captured["command"] = command
+            captured["stdin_data"] = stdin_data
+            proc = subprocess.Popen(
+                ["/bin/true"],
```
sukanya1426: /bin/true doesn't exist on macOS (only /usr/bin/true), so this is the one test that fails on a Mac dev box, it actually executes the path instead of just passing it as a string. Fine on CI/Linux. shutil.which("true") would make it portable. Not a blocker though.

munimthahmid (approved):
@Qian-Cheng-nju  I did a thorough review, and everything seems to be very well implemented and working perfectly. I also did some one-shot agent runs, and they passed. I found no blocking  issues, and I’m approving these changes.

sukanya1426 (changes requested):
Everything seems working well. Also verified a live litellm_oneshot call against OpenAI. The test suite is green on Linux/CI, the only local issue I hit is the /bin/true nit below. I didn't run copilot_oneshot live since it requires a token. LGTM.

tests/evaluator/test_oneshot_runner_regressions.py
Qian-Cheng-nju: Fixed in 4b7ab14 by resolving the executable with `shutil.which("true")` before launching it. The targeted regression test passes, and the full suite remains green. Thanks!

sukanya1426 (approved):
LGTM....

PR #66, opened 2026-07-19
Add structured Copilot usage telemetry

Qian-Cheng-nju:
## Summary

- Add provider-neutral structured usage records and aggregate them across subagents, retries, and continuations.
- Collect Copilot usage from official CLI OpenTelemetry and SDK `assistant.usage` events.
- Run agentic Copilot through the supported `-p` mode and force an isolated file exporter for every task.
- Keep unavailable token counts as `null`, explicit zeroes as exact zeroes, and provider-native costs in their reported units.


src/evaluator/backends/copilot.py
```diff
@@ -5,11 +5,315 @@
...
+    return end - start
+
+
+def _copilot_costs(attributes: dict[str, Any]) -> tuple[UsageCost, ...]:
+    costs: list[UsageCost] = []
+    monetary = nonnegative_float(attributes.get("github.copilot.cost"))
+    if monetary is not None:
+        # GitHub documents this as monetary cost but does not publish a currency
+        # on the span, so preserve the provider's unit instead of assuming USD.
+        costs.append(UsageCost(monetary, "provider_monetary", "github.copilot.cost"))
```
munimthahmid: This unit name is a bit misleading. GitHub's OTel docs call `github.copilot.cost` "monetary cost," but its SDK docs describe the corresponding usage cost as the "model multiplier cost for billing." I did a  Copilot run with Claude Haiku 4.5, this value was `0.33`, matching the model's `0.33` multiplier, while AI-unit usage was reported separately through `nano_aiu`. Could we use `model_multiplier` instead of `provider_monetary`?

munimthahmid (changes requested):

src/evaluator/backends/copilot.py
Qian-Cheng-nju: Updated `github.copilot.cost` to use the `model_multiplier` unit, matching the SDK `assistant.usage.cost` semantics and the observed CLI value.

sukanya1426 (approved):
Everything looks good and works as described.

One more non blocking question is that , in _authoritative_costs, we sum github.copilot.cost across all invoke_agent roots, including nested sub-agents. That seems correct as long as each span's cost is just its own turn cost and not cumulative of its children. The logic and tests look right to me, I just don't have a real multi-agent run to compare the numbers against. Have you been able to verify this against actual usage?

munimthahmid (approved):
@Qian-Cheng-nju  Reviewed the implementation thoroughly, and the new structured usage telemetry worked great in my tests. Did some Copilot and one-shot LiteLLM calls, and the token usage, request count, model details, and completeness status were reported correctly for both. Copilot's cache, reasoning, and cost details were also reported correctly. Approving.

Qian-Cheng-nju:
Thanks for raising this. I verified with Copilot CLI 1.0.71 that each invoke_agent root only includes its own direct calls, while sub-agents have separate roots. So summing the roots does not double-count usage. Merged

PR #67, opened 2026-07-19
Add optional reasoning effort support

munimthahmid:
## Summary

- Add optional `--reasoning-effort` support across compatible backends.
- Preserve existing backend defaults when omitted.
- Forward the selected effort directly to the backend.
- Record the requested effort in result metadata.

## Validation

- Live Codex runs passed at `low` and `medium`.
- Live Copilot + Claude Haiku 4.5 passed at `medium`.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -59,6 +59,14 @@ uv run tlaps-bench run --backend litellm --model claude-sonnet-4-6
 uv run tlaps-bench run --backend litellm_oneshot --model claude-sonnet-4-6
 ```
 
+### Reasoning effort
+
+Use the optional `--reasoning-effort` flag to override a model's reasoning budget:
+
+```bash
+uv run tlaps-bench run --backend codex --model gpt-5.6-sol --reasoning-effort low
+```
+
 ### Authentication
 
 The runner passes credentials into the container in two ways:
@@ -139,6 +147,7 @@ uv run tlaps-bench run [flags]
 | `--backend` | `codex` | Evaluator backend to use |
 | `--mode` | `proof-completion` | Benchmark mode |
 | `--model` | (backend default) | Override the model |
+| `--reasoning-effort` | (backend behavior) | Pass a backend/model-specific reasoning effort; omitted preserves existing behavior |
 | `--filter` | (all benchmarks) | Substring match on path, comma-separated |
 | `--jobs` | `1` | Number of parallel backend runs |
 | `--timeout` | `28800` | Per-benchmark backend timeout in seconds |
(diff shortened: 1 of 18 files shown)


Saadmrp1038 (changes requested):
@munimthahmid where is pi support 👀
Some other issues:
1. Preflight passed with invalid reasoning effort value.
```bash
uv run tlaps-bench run   --backend litellm   --model claude-sonnet-4-6   --reasoning-effort definitely-invalid   --filter GCD_GCD3
Container mode: ON (image: tlaps-bench-base)
Preflight: validating 'litellm' (install + auth + model + firewall)...
Preflight: OK
Backend: litellm (model=claude-sonnet-4-6)
Effort:  definitely-invalid
Mode:   proof-completion — Proof completion — fill in the last theorem's PROOF OBVIOUS
Output:  /home/pial/Projects/tlaps-bench/results/proof-completion/litellm/20260719_141052
Found 1 benchmarks
```
2. Takes in empty value. Records as empty but the code in `claude_code.py` defines `self.reasoning_effort or "max"`
```
uv run tlaps-bench run   --backend claude_code  --model claude-sonnet-4-6   --reasoning-effort=  --filter GCD_GCD3
Container mode: ON (image: tlaps-bench-base)
Preflight: validating 'claude_code' (install + auth + model + firewall)...
Preflight: OK
Backend: claude_code (model=claude-sonnet-4-6)
Effort:  
Mode:   proof-completion — Proof completion — fill in the last theorem's PROOF OBVIOUS
Output:  /home/pial/Projects/tlaps-bench/results/proof-completion/claude_code/20260719_141922
Quota:   gate OFF — usage probe returned no data (API-key auth or no subscription usage to read)
Found 1 benchmarks
```

munimthahmid:
@Saadmrp1038 Sorry, missed Pi. Added now.

For model validation, initially I thought we could leave it up to the user since reasoning effort can vary a lot by model. Now I’ve added the list of reasoning efforts accepted by each backend.

So there is still that minor issue. For example, GPT-5.5 doesn’t support `ultra`, but GPT-5.6 Sol does. So Codex accepts `ultra`, but an unsupported model/effort combination will fail in preflight. I don’t see a clean way to validate this without maintaining model-specific lists, which feels a bit too hardcoded. Let me know if you have any suggestions.

munimthahmid:
Also, it seems like LiteLLM preflight has a bug. From your run, even though LiteLLM returned an error for the invalid reasoning effort, preflight still passed. I also tested it with an invalid model name, and it still passed because the LiteLLM agent exits with code 0, which our preflight considers a success.

```console
$ uv run tlaps-bench run --backend litellm --model definitely-invalid --filter GCD_GCD3 --infra-retries 0 --timeout 60
Container mode: ON (image: tlaps-bench-base)
Preflight: validating 'litellm' (install + auth + model + firewall)...
Preflight: OK
Backend: litellm (model=definitely-invalid)
```

Saadmrp1038 (approved):

PR #68, opened 2026-07-20
Add usage and cost telemetry for Claude Code and LiteLLM #68

sukanya1426:
Summary

- Add structured usage and cost telemetry for the Claude Code and LiteLLM backends.
- Record per-request tokens and native USD cost in a usd unit, separate from Copilot's model_multiplier.
- Keep unavailable values null and mark partial runs as a lower bound instead of zero.
- Dedupe Claude Code streamed events and take settled output and cost from the result event.

Validation

- Live Claude Code runs passed single-turn, multi-turn, and on timeout.
- Live LiteLLM gpt-4o-mini run passed with the exact provider reported cost.

Notes

- Builds on #66, extends its usage framework (usage.py, parse_usage) to Claude Code and LiteLLM.
- The LiteLLM preflight/exit-code fix comes from #67 on rebase, so it is not re done here.


sukanya1426:
@Saadmrp1038  Could you please review this PR? Thank you.

Saadmrp1038 (approved):

PR #69, opened 2026-07-20
Allow custom OpenAI-compatible endpoint through firewall

lemmy:
The agent container firewall only whitelisted a fixed list of known LLM provider hosts, so OpenAI-compatible endpoints reached via a custom base URL (e.g. SomeCompany's inference API) were blocked despite OPENAI_API_BASE / OPENAI_BASE_URL already being forwarded into the container.

Derive the host from OPENAI_API_BASE / OPENAI_BASE_URL and add it to the firewall allow-list, mirroring the existing Azure OpenAI handling. This lets `litellm --model openai/<model>` target any OpenAI-compatible endpoint via env alone. Document the workflow in the LiteLLM section of docs/USAGE.md.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -59,6 +59,22 @@ uv run tlaps-bench run --backend litellm --model claude-sonnet-4-6
 uv run tlaps-bench run --backend litellm_oneshot --model claude-sonnet-4-6
 ```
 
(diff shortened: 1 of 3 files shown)


tianyin (approved):
LGTM

PR #70, opened 2026-07-21
Add reliable Codex and Pi usage telemetry

munimthahmid:
## What changed

- Add provider-native usage telemetry for Codex `turn.completed` events and Pi `message_end` events, normalized into the structured usage schema while preserving the existing top-level token fields.
- Keep Codex child agents enabled and aggregate their native rollout token deltas without double-counting inherited parent usage.
- Aggregate usage across retries and continuation rounds, while retaining trustworthy lower bounds for incomplete or damaged streams.
- Prevent retries when a request may already have reached the provider, and retain Pi's native per-request metadata and estimated USD costs.
- Pin the validated Codex `0.144.6`, Pi `0.80.10`, and Kiro provider `0.8.1` versions.

## Testing

- `uv run pytest -q` (889 passed, 3 skipped)
- `uv run ruff check .`
- `uv run ruff format --check .`
- `docker buildx build --check --file docker/base.Dockerfile .`
- Verified live requests through Codex and Pi, including a `gpt-5.6-sol` high-reasoning Codex session with a real child agent and complete aggregated usage.


munimthahmid:
@Saadmrp1038, if possible, can you review this PR as well? Thanks a lot.

Qian-Cheng-nju:
@sukanya1426 Could you review that? I think Pial is very busy recently

sukanya1426:
@Qian-Cheng-nju sure!

src/evaluator/runner.py
```diff
@@ -454,8 +544,11 @@ def _run_once(workspace=workspace, canonical_dir=canonical_dir):
             )
             result["termination_reason"] = classify(ctx)
 
-            # A genuine attempt (any output tokens) is never re-run.
-            infra_retriable = result["termination_reason"] == TerminationReason.INFRA_ERROR and output_tokens == 0
+            # Structured telemetry can prove a model call happened even when the
+            # legacy transcript stream contains zero output tokens.
+            infra_retriable = result["termination_reason"] == TerminationReason.INFRA_ERROR and not (
```
sukanya1426: main added `backend.is_infra_retryable(ctx)` to this same expression, and one-shot backends override it. If your line just replaces main's, they lose that gate and become replayable when they shouldn't be.

ANDing the two doesn't work either, your helper returns True whenever model_requests > 0, but the one-shot protocol says a request marked `retryable: true` with zero output IS safe to replay. I checked with both gates, test_copilot_oneshot_retries_zero_output_infra_then_succeeds and test_copilot_retry_result_does_not_keep_prior_attempt_wire_metadata both fail, drop your clause and all 27 one-shot tests pass.

so,I think this needs an explicit precedence rule.

src/evaluator/usage.py
```diff
@@ -333,20 +358,50 @@ def from_requests(
...
-            validation_warnings.extend(
-                f"{field} is missing for some model requests; total is a lower bound" for field in partial_fields
+        core_fields_missing_everywhere: list[str] = []
+        empty_exactness_errors: list[str] = []
+        if complete and request_tuple:
+            for field in ("input_tokens", "output_tokens"):
+                values = [getattr(request, field) for request in request_tuple]
+                if field not in provider_total_fields and all(value is None for value in values):
+                    core_fields_missing_everywhere.append(field)
+        elif complete and not request_tuple:
```
sukanya1426: The branch currently treats missing per-request records as a 'free run.' However, Claude Code can report only aggregate totals (for example, num_turns=4, 30 input / 90 output) without any per-request events, and that's still a valid complete summary. With this change, those runs get marked as incomplete and receive a warning about zero tokens, even though it doesn't apply to them.

The zero-exactness check should only cover actual zero-cost claims. Aggregate only runs just need authoritative core totals, so gate the existing block on `derived["model_requests"]` being falsy, and in the truthy branch append to `core_fields_missing_everywhere` when a core total isn't authoritative.

I ran that locally, it restores `status: complete, model_requests: 4` for the Claude Code case, and all your exactness tests stay green.

sukanya1426 (changes requested):
Also, could you resolve the merge conflicts ?

Qian-Cheng-nju:
Fixed a small issue, thanks! LGTM

PR #71, opened 2026-07-21
Add proof-from-scratch evaluator contract

munimthahmid:
> [!IMPORTANT]
> This remains a draft until the proof-from-scratch generator emits `manifest.json` and the editable-region markers. Runs fail closed without that producer-side contract.

## What changed

- Make the manifest the sole source of proof-from-scratch task discovery and context, with strict validation and no heuristic fallback.
- Restrict agent edits to marked helper and proof regions, preserve the remaining scaffold byte-for-byte, and reject forbidden or unproved helper declarations.
- Keep context files read-only during execution and grade the submitted target against a fresh canonical context that was never exposed to the agent.
- Apply the same contract to one-shot responses and continuation rounds, and update the prompts and usage documentation.


munimthahmid:
@sukanya1426  Can you also review this PR when you get a chance? Since #80  depends on it, we should make sure both PRs stay in sync.

sukanya1426:
Sure.

src/common/proof_from_scratch_contract.py
```diff
@@ -0,0 +1,360 @@
...
+        for context_key, context_path in context_entries:
+            if context_key == task_key or context_path == task_path:
+                raise ManifestError(f"manifest task {task_key!r} includes itself in its context")
+            if context_key in task_keys or context_path in task_paths:
+                other_task = task_paths.get(context_path, context_key)
+                raise ManifestError(f"manifest task {task_key!r} includes task {other_task!r} in its context")
+            context_paths.append(context_path)
+
+        resolved_paths = tuple(context_paths)
+        _validate_module_files(task_key, ((task_key, task_path), *context_entries))
```
sukanya1426: The validator checks paths, symlinks, duplicate basenames, and module names, but it never verifies that the context covers the task's full EXTENDS/INSTANCE closure. I tested this by removing a context module, and the run ended up as [SANY-INVALID] / GATES-FAILED: A, identity, so a generator bug gets scored as agent tampering. Since the manifest is now the only source of context, it should validate the closure here and raise ManifestError instead.

src/common/proof_from_scratch_contract.py
```diff
@@ -0,0 +1,360 @@
...
+        boundaries[task_key] = TaskBoundary(
```
sukanya1426, line 332: Another small improvement: call parse_editable_regions while loading tasks. Missing markers currently slip through loading and only surface as an exit-3 ERROR during grading. Better to catch them early.

src/evaluator/runner.py
```diff
@@ -999,16 +1057,38 @@ def _run_continuations(
             # never got to work on (quota cap or 0-token startup death).
             cut_short = run.quota_exhausted or run.infra_retriable
             if not cut_short:
+                grading_canonical_dir = run.canonical_dir
+                if getattr(mode, "canonical_replay_required", False):
+                    grading_canonical_dir = _make_canonical_dir(
```
sukanya1426: This snapshot is taken after the agent finishes and reads from item.benchmark_path in the repo. In --no-container mode the agent has host write access, so it can edit the source and the "fresh" snapshot picks up the tampered bytes, which also feeds get_baseline_text, so nothing independently catches it. Take it once before _run_backend_with_retries and reuse it for grading.

src/common/check_proof.py
```diff
@@ -491,6 +493,61 @@ def resolve_benchmark_dir(explicit, filepath, target_name):
...
+
+    dependency_dir = benchmark_dir if require_canonical else os.path.dirname(filepath)
+    for dep_file in sorted(glob.glob(os.path.join(dependency_dir, "*.tla"))):
+        dep_basename = os.path.basename(dep_file)
+        if dep_basename != basename:
+            shutil.copy2(dep_file, os.path.join(destination, dep_basename))
+    return staged_file
+
+
+def check_editable_region_integrity(filepath, benchmark_dir):
```
sukanya1426, line 520: Byte exact comparison means even a trailing newline at EOF or a CRLF rewrite gets flagged as CHEAT-DETECTED, scaffold_unchanged (gate A). I confirmed both cases. _reconstruct_marked_solution already treats these as harmless retyping for one-shot, but agentic backends don't. We should either normalize line endings and trailing newlines before comparing, or give these their own non-cheat reason code. Otherwise, editor noise will inflate the reported cheating rate.

src/evaluator/modes/proof_from_scratch.py
```diff
@@ -9,9 +9,62 @@
...
+    def is_benchmark_file(self, path: str) -> bool:
+        """Return whether ``path`` is a task declared by the manifest."""
+
+        try:
+            resolved = Path(path).resolve(strict=True)
+        except (OSError, RuntimeError):
+            return False
+        return resolved in self._boundaries_by_path
+
+    def get_benchmark_files(self, filter_pattern: str | None = None) -> list[str]:
```
sukanya1426, line 45: main() calls this at runner.py:1721 with no try/except, so a missing or malformed manifest escapes as a raw ManifestError traceback and never reaches the parser.exit(2, ...) handling on the next line. Catch it in main() or handle it here, so the user gets an error message instead of a stack trace.

sukanya1426 (changes requested):

src/evaluator/runner.py
munimthahmid: Good catch. Fixed. In --no-container mode, we now save the original target and context files in memory before starting the agents. I

src/common/check_proof.py
```diff
@@ -491,6 +493,76 @@ def resolve_benchmark_dir(explicit, filepath, target_name):
...
+        return [(0, f"cannot read submitted proof-from-scratch task: {exc}", "SCAFFOLD_MODIFIED")]
+
+    try:
+        submitted_regions = parse_editable_regions(submitted_source)
+    except EditableRegionError as exc:
+        return [(0, f"editable-region markers were modified: {exc}", "SCAFFOLD_MODIFIED")]
+
+    if submitted_regions.fixed_segments != canonical_regions.fixed_segments:
+
+        def normalize_format(regions):
```
sukanya1426, line 559: One thing I noticed is that this only handles a single trailing newline, so the common case of an editor adding a newline still gets reported as SCAFFOLD_MODIFIED/CHEAT-DETECTED. Switching to suffix = suffix.rstrip("\n") fixes that without affecting real scaffold modifications.

munimthahmid: Fixed. Extra newlines at EOF are now ignored.

sukanya1426 (approved):
Seems alright now. Thanks for addressing the issues. Over to @Qian-Cheng-nju , can you please take a look and validate ?

Qian-Cheng-nju:
LGTM!

PR #77, opened 2026-07-24
Add Cursor CLI backend

lemmy:
Add a `cursor` backend running `cursor-agent` non-interactively (--print, stream-json, --force, --sandbox disabled). Auth uses the mounted `cursor-agent login` token (~/.config/cursor/auth.json) or CURSOR_API_KEY.

Cursor's API/agent endpoints resolve to large, rotating AWS IP pools that the pin-at-start firewall cannot track, so add an opt-in `unrestricted_https_egress` backend flag that switches the container firewall to allow all outbound HTTPS (443) while still blocking other egress. Also generalize credential mounts with a `source_dir` since Cursor's token lives outside ~/.<name>.

### DIFF
--- a/docker/install-scripts/install-cursor.sh
+++ b/docker/install-scripts/install-cursor.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+set -euo pipefail
+# Cursor Agent CLI ships as a self-contained binary installed under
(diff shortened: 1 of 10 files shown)


lemmy:
My agent attempted a dynamic whitelist, but it failed due to an excessive number of IPs.

PR #80, opened 2026-07-27
Add proof-from-scratch layered generation

sukanya1426:
> [!IMPORTANT]
> This is the producer side of #71. It does not regenerate `benchmark/proof-from-scratch/` in place, that should land after #71 merges, since the evaluator still discovers tasks by filename heuristic today.

## What changed

- Split each task into a read-only shared model, a read-only per-target definitions module, and an editable task module holding only the theorem scaffold and the agent's helper and proof regions.

- Emit a `manifest.json` giving every task its exact read-only context, so no task inherits a sibling's definitions and the evaluator never has to infer context from siblings.

- Keep original-proof definitions and module-level `USE`/`HIDE` hints out of the read-only layers, and audit each generated layer for leaked proof artifacts and shared ownership.

- Give pure lemma targets a self-contained definitions layer instead of the shared model, carrying only the declarations they use.

## Validation

 Validated across 293 generated tasks, all load and parse under #71's manifest loader and region parser, all SANY-parse with only their manifest context, no proof artifacts leak, no `Defs` is shared, and an honest proof with a helper lemma passes `tlapm`.

## Gaps

17 existing tasks (`OpenAddressing`, `etcd_raft`, `GermanProtocol`, `EWD687a_proof`) have no `source/` counterpart and can't be generated, they need sources or hand-authored layers. Task count also depends on --allow-no-proof: 293 with it, against 227 in the current dataset.


sukanya1426:
@munimthahmid Generator side is working, a few things I had in mind, 

- Marker strings are defined twice now, mine in generate.py, yours in proof_from_scratch_contract.py. They have to stay byte-identical or every task fails your scaffold check. Do you want the generator to just import them from your module?

- Context isn't always Model + Defs, 151 of 293 tasks are pure lemma targets with no model layer at all, just a Defs. A few have a third file (Paxos_Refinement also needs Consensus.tla via an INSTANCE). Your loader handles it, but I think we should check the prompt wording and read-only mounting don't assume exactly two always.

- One odd case with USE. I'm stripping module-level USE/HIDE from the read-only layers like we said, but three sources do USE NAssumption a fact, not a DEF  and your helper region only takes DEF directives. So there's nowhere for those to go and the agent would just have to write BY NAssumption. That's fine by me, only mentioning it so it doesn't look like a bug later.

- Seems like unnamed theorems are common. Your example had THEOREM Safety == ..., but plenty of real tasks are just THEOREM Spec => []MutualExclusion. Does your canonical-statement check assume a name?

src/dataset/proof_from_scratch/generate.py
```diff
@@ -719,6 +915,116 @@ def compute_sibling_deps(targets):
...
+                f"{sorted(defs_set & artifacts)}\n"
+            )
+
+        statement_text = strip_comments(_statement_text(target_thm, source_lines)).strip()
+        task_text = build_task_module(bench_module_name, defs_module, statement_text)
+        task_path = os.path.join(out_dir, f"{bench_module_name}.tla")
+        with open(task_path, "w", encoding="utf-8") as f:
+            f.write(task_text)
+
+        dep_basenames = copy_deps(dump, source_path, out_dir, reachable)
```
munimthahmid: This still copies the whole dependency file, so the read-only context can contain old theorems and proof helpers. That defeats the purpose of proof-from-scratch.

For example, the Voting context contains `AllSafeAtZero` with `PROOF OMITTED`, so `BY AllSafeAtZero` passes as the task proof.

We should generate a cleaned dependency file for each task containing only the definitions, constants, variables, and assumptions it needs. Old theorem and lemma declarations should be removed completely.

src/dataset/proof_from_scratch/generate.py
```diff
@@ -1002,23 +1345,123 @@ def main():
...
+            removed = cross_dir_dedup(generated_paths, audit_writer)
+            # Input SANY gate: every emitted task must parse under standalone
+            # tla2sany. Flags failures (manifest + audit log); does not drop.
+            sany_gate(output_root, audit_writer=audit_writer, label="sany-gate-l2")
+            # Triviality gate: a task whose PROOF OBVIOUS placeholder already
+            # verifies is degenerate (a no-op submission would PASS grading).
+            dropped = len(
+                triviality_gate(output_root, audit_writer=audit_writer, label="triviality-gate-l2", drop=True)
+            )
+
```
munimthahmid, line 1703: SANY and triviality checks should not be skipped. We need to update them to use each task’s exact context from the manifest in layered mode.

src/dataset/proof_from_scratch/generate.py
```diff
@@ -719,6 +915,116 @@ def compute_sibling_deps(targets):
...
+        defs_module = f"{bench_module_name}Defs"
+        keep_decls = not use_model
+
+        defs_text = build_defs(source_lines, dump, defs_set, defs_module, model_module, keep_decls)
+        defs_path = os.path.join(out_dir, f"{defs_module}.tla")
+        with open(defs_path, "w", encoding="utf-8") as f:
+            f.write(defs_text if defs_text.endswith("\n") else defs_text + "\n")
+        if _THEOREM_SCAN.search(defs_text):
+            audit_writer.write(f"[audit] {source_path}: LEAK Defs {defs_module} contains a THEOREM/LEMMA\n")
+        if defs_set & artifacts:
```
munimthahmid: defs_set and artifacts can never have anything in common by construction, so this condition will always be false. We should check the definitions actually written to the generated context files instead.

munimthahmid (changes requested):
@sukanya1426 I found some issues with the implementation. Can you check and fix them?

munimthahmid:
> Marker strings are defined twice now, mine in generate.py, yours in proof_from_scratch_contract.py. They have to stay byte-identical or every task fails your scaffold check. Do you want the generator to just import them from your module?

@sukanya1426 You can keep your copies for now. After this PR is merged, I will rebase #71 and make the generator import the shared constants.


> Context isn't always Model + Defs, 151 of 293 tasks are pure lemma targets with no model layer at all, just a Defs. A few have a third file (Paxos_Refinement also needs Consensus.tla via an INSTANCE). Your loader handles it, but I think we should check the prompt wording and read-only mounting don't assume exactly two always.

No, we don't assume exactly two files. The prompt and read-only mounting handle any number of context files.

> One odd case with USE. I'm stripping module-level USE/HIDE from the read-only layers like we said, but three sources do USE NAssumption a fact, not a DEF and your helper region only takes DEF directives. So there's nowhere for those to go and the agent would just have to write BY NAssumption. That's fine by me, only mentioning it so it doesn't look like a bug later.

This is fine. The agent can use NAssumption inside the proof with BY NAssumption or a proof-level USE NAssumption.

> Seems like unnamed theorems are common. Your example had THEOREM Safety == ..., but plenty of real tasks are just THEOREM Spec => []MutualExclusion. Does your canonical-statement check assume a name?

Unnamed target theorems are fine. The checker does not require the target theorem to have a name.

sukanya1426:
@munimthahmid Thanks for the feedbacks! Fixed them, can you please take a look again ?

I also ran into another issue. TwoPhase_proof_line17 isn't really self-contained, its statement is just THEOREM Implementation, which points to THEOREM Implementation == Spec => A!Spec in TwoPhase.tla. This worked before because the dependencies still contained their theorems, but now the context modules don't, so the reference can't be resolved and the task gets dropped by the SANY gate.

Would you prefer that I inline the actual statement (Spec => A!Spec) so the task is self-contained, or should I just retire it?

src/dataset/proof_from_scratch/generate.py
```diff
@@ -719,6 +1071,142 @@ def compute_sibling_deps(targets):
...
+
+        statement_text = strip_comments(_statement_text(target_thm, source_lines)).strip()
+        task_text = build_task_module(bench_module_name, defs_module, statement_text)
+        task_path = os.path.join(out_dir, f"{bench_module_name}.tla")
+        with open(task_path, "w", encoding="utf-8") as f:
+            f.write(task_text)
+
+        # Dependencies are written once at finalize, pruned to the union of what
+        # every task using them mentions — a per-task copy is impossible because
+        # they share one filename, and the module name must match the filename.
```
munimthahmid: Dependency files are still shared between tasks. This means a task can see definitions that only another task needs. Each task should get its own cleaned dependency files. The same filename is fine because each task runs in its own workspace.

src/dataset/proof_from_scratch/generate.py
```diff
@@ -1002,23 +1533,269 @@ def main():
...
+    failures = []
+    with ThreadPoolExecutor(max_workers=jobs) as ex:
+        futs = [ex.submit(check, it) for it in manifest.items()]
+        for fut in as_completed(futs):
+            task_key, ok, err = fut.result()
+            if not ok:
+                failures.append((task_key, err))
+                audit_writer.write(f"[audit] {task_key}: FAILED standalone SANY with its manifest context — {err}\n")
+
+    # Unlike the suite-wide gate this one DROPS. In layered mode the manifest is
```
munimthahmid: SANY failures should not remove tasks from the manifest because the existing SANY check only flags failed tasks, doesn't remove them. I think layered mode should do the same.

src/dataset/proof_from_scratch/generate.py
```diff
@@ -1002,23 +1533,269 @@ def main():
...
+
+    from dataset.triviality_audit import check_task, find_tlapm, find_tlapm_lib
+
+    tlapm_path = find_tlapm()
+    tlapm_lib = find_tlapm_lib(tlapm_path) if tlapm_path else None
+    if not tlapm_path or not tlapm_lib:
+        msg = "tlapm not found — layered triviality gate SKIPPED (degenerate tasks may ship)"
+        audit_writer.write(f"[audit] {msg}\n")
+        print(f"⚠️  [layered-triviality-gate] {msg}")
+        return []
```
munimthahmid: We should make sure that degenerate tasks are not generated. So missing TLAPM should make generation fail. That is the current behavior as well. The only exception is when someone passes --skip-gates. Then we skip the triviality and SANY gates.

munimthahmid (changes requested):
@sukanya1426  The previous issues are partly fixed, but the updated implementation still needs some changes. I left inline comments.

sukanya1426:
@munimthahmid Thanks! I've finished the changes on my end. I still need your call on TwoPhase_proof_line17, though. It's now being flagged instead of dropped, as you requested, so it stays in the manifest but still can't parse. That means every submission for it fails. Should we inline Spec => A!Spec, or retire it?

munimthahmid:
@sukanya1426 Thanks, I am reviewing it. And yes, about the TwoPhase problem, I checked, and it seems like TwoPhase_proof_line17 is just another proof of the existing Implementation theorem in TwoPhase.tla. So we should remove the generated task. But we need to keep the original source file unchanged. 

Sorry, I forgot to mention it earlier.

src/dataset/proof_from_scratch/generate.py
```diff
@@ -933,6 +1479,70 @@ def process_file(
...
+    """Targets for examples whose sources live under the benchmark tree.
+
+    Not every example is mirrored under `source/`: OpenAddressing and etcd_raft
+    were committed straight into `benchmark/proof-from-scratch/`, so a scan of
+    `source/` alone silently loses those tasks. Their directories are treated as
+    source too.
+
+    The files are staged into a temp directory first because the benchmark tree
+    is also the output tree — generating in place would overwrite a module while
+    a later task still needs to read the original.
```
munimthahmid: We should not use generated benchmark files as source files. The original files are available upstream:

- [`etcd_raft`](https://github.com/etcd-io/raft/tree/main/tla)
- [`OpenAddressing.tla`](https://github.com/lemmy/Examples/blob/mku-OA/specifications/TLC/OpenAddressing.tla)
- [`OpenAddressing_proof.tla`](https://github.com/lemmy/Examples/blob/mku-OA/specifications/TLC/OpenAddressing_proof.tla)

We should add the needed files under `source/` and generate these tasks the same way as the other tasks. Otherwise the first run overwrites the input, and the next run loses these tasks.

cc @Qian-Cheng-nju   can you confirm that these are the original sources?

munimthahmid (changes requested):

src/dataset/proof_from_scratch/generate.py
```diff
@@ -496,6 +524,185 @@ def copy_deps(dump, source_path, out_dir, reachable):
...
+            body = strip_comments("\n".join(lines[o["loc"]["line_start"] - 1 : o["loc"]["line_end"]]))
+            refs |= _scan_identifiers(body)
+            adj.setdefault(o["name"], set()).update(refs)
+        for i in d.get("instances", []):
+            if not i.get("name"):
+                return None
+            adj.setdefault(i["name"], set()).update(i.get("references", []))
+            # An INSTANCE statement is never deleted from a dependency, so every
+            # name in its WITH substitution stays live — `WITH chosen <- chosen`
+            # keeps `chosen` even when no task mentions it.
```
munimthahmid: Returning `None` here keeps the whole dependency group. We should keep the unnamed `LOCAL INSTANCE` line unchanged and continue removing other definitions the task does not need.

It is very important that each task only gets the definitions it needs, since that is the whole purpose of this split.

src/dataset/proof_from_scratch/generate.py
```diff
@@ -1002,23 +1642,268 @@ def main():
...
+            if task_key == keeper:
+                continue
+            del manifest[task_key]
+            os.remove(os.path.join(output_root, task_key))
+            audit_writer.write(
+                f"[audit] {task_key}: identical task and context to {keeper} — removed (filter C, cross-dir dedup)\n"
+            )
+            removed += 1
+
+    sweep_unreferenced_context(output_root, manifest, audit_writer)
```
munimthahmid: When running with `--filter`, this function deletes every other TLA+ file in the output directory. So generating one group deletes all other existing tasks, and the full manifest is also replaced with the filtered one. A filtered run should only touch the selected tasks and leave the rest of the dataset unchanged.

munimthahmid (changes requested):
@sukanya1426  Thanks for the updates. The previous  issues are fixed. I found a few new issues in the updated code and left inline comments. Could you please check and fix them?

src/dataset/proof_from_scratch/generate.py
munimthahmid: @sukanya1426 you can go ahead and add the sources. We can update it later if Qian suggests any changes

sukanya1426:
@munimthahmid  The reviews were really on point xd. Thanks for catching those. I’ve fixed them all. Mind taking another look? :3

sukanya1426:
@munimthahmid  The dataset now regenerates 232/233 tasks exactly. The only exception is `TwoPhase_proof_line17`, which I removed as discussed.

Just noting this since the target was to preserve all 233 tasks. I think removing it is the cleaner option because it only mirrors `TwoPhase.tla`'s `Implementation` theorem and can't work as a standalone goal anymore. If we want to keep 233, we’d need to inline `Spec => A!Spec` instead.

src/dataset/proof_from_scratch/generate.py
```diff
@@ -1002,23 +1569,299 @@ def main():
...
+    if run_gates:
+        layered_sany_gate(output_root, manifest, audit_writer)
+        dropped = layered_triviality_gate(output_root, manifest, audit_writer)
+
+    # Untouched tasks keep their stored entry; a task this run regenerated is
+    # represented only by what survived dedup and the gates.
+    complete = {k: v for k, v in existing.items() if k not in regenerated}
+    complete.update(manifest)
+    sweep_unreferenced_context(output_root, complete, audit_writer)
+
```
munimthahmid: Filtered generation is still broken. I ran `--filter TwoPhase` on an existing layered dataset, but `TwoPhase_proof_line17` stayed in the manifest and on disk even though it should be removed. We should remove the old tasks for the selected source first, then add the newly generated tasks.

Also, if the output directory is not empty but has no valid manifest, the filtered run should stop before changing anything. Otherwise, it treats the filtered tasks as the entire dataset and deletes all other tasks.

munimthahmid (changes requested):

src/dataset/proof_from_scratch/generate.py
sukanya1426: Both issues are fixed now.

The removal bug was caused by using the current run output as the source of truth. Since `TwoPhase_proof_line17` was filtered out earlier, it was never included there and the old entry stayed. Now we remove based on regenerated sources, old tasks are cleared first, then the new output is added.

Also fixed the `TwoPhase.tla` / `TwoPhase_proof.tla` prefix overlap by using the longest matching source basename, so tasks don't get removed incorrectly.

For the second issue, partial runs now fail early if `.tla` files exist but the manifest is missing, unreadable, or invalid. Empty directories still work normally. Added regression tests for all of these cases.

munimthahmid:
@sukanya1426 The generator changes look good now. I’ll test #71 and #80 together locally. It would be great if you could do the same and run 3–5 real agent attempts on different task types.

If everything works, we’ll merge #71  first. Then you will need to rebase #80  onto `main` and import the shared marker constants from #71  before merging #80.

Once both PRs are merged, we can regenerate the dataset and add `manifest.json` in a follow-up PR.

sukanya1426:
@munimthahmid sounds good. I will let you know after testing them locally.

munimthahmid:
@Qian-Cheng-nju  @sukanya1426 I tested #71 and #80 together locally. Here are the results.

### Dataset generation

- Default generation produced 203 candidates. It removed 14 duplicates and 6 trivial tasks, leaving 183 tasks in `manifest.json`.
- Generation with `--allow-no-proof` produced 340 candidates. It removed 25 duplicates and 9 trivial tasks, leaving 306 tasks in `manifest.json`.
- Both runs completed without any generation errors, SANY failures, dependency leaks, or missing task/context files.

### Real agent tests

I ran 5  agent attempts across different task types and model/reasoning combinations:

- `gpt-5.6-sol` with high reasoning passed `sums_even_T1`, `Allocator_InitMutex`, and `Paxos_Refinement`.
- `gpt-5.5` with medium reasoning passed `sums_even_T1`.
- `OpenAddressing_Contains` did not pass. Upstream only defines `Contains`, but our benchmark turns it into a theorem. The agent found a counterexample, so the evaluator correctly returned `FAIL`. However, we should not count this as an agent failure because the theorem itself is invalid.

The agents only changed the allowed marker regions. The evaluator used the exact context from `manifest.json` and mounted the context files as read-only. So the evaluator and generator are working together correctly.

### Remaining issue

- Current dataset: 233 tasks.
- Default generation: 183 tasks because it skips the existing no-proof groups.
- Global `--allow-no-proof`: 306 tasks because it adds every no-proof target.

It seems the current benchmark includes some no-proof tasks. We can either add a vetted allow-list so they are included by default, or keep only the 183 tasks with source proofs. Which way should we go?

sukanya1426:
@Qian-Cheng-nju @munimthahmid I tested #71 and #80 together locally too. 

Dataset generation

- Default produced 203 candidates, removed 14 duplicates and 7 trivial, leaving 182 tasks. With --allow-no-proof: 340 candidates, 25 duplicates and 9 trivial removed, leaving 306. Both runs clean.
- The one difference from Munim's numbers is default mode, 182 vs his 183. That's TwoPhase_proof_line17.

Real agent tests

- I ran 5 agent attempts with gpt-5.5 via litellm across different task shapes, all passed: Cantor1_cantor, BubbleSort_IsPermOfTransitive, SimpleMutex_Safety, Peterson_MutualExclusion, and EWD840_TerminationDetection (up to 125 obligations).
- The agents only changed the marker regions, and the evaluator used exactly the manifest context, Cantor1_cantor got only its own Defs despite 9 sibling Cantor tasks in the same directory. I also checked the failure side, placeholder, ASSUME FALSE in helpers, PROOF OMITTED, and a weakened statement each fail with the right gate and reason code.

Qian-Cheng-nju:
> It seems the current benchmark includes some no-proof tasks. We can either add a vetted allow-list so they are included by default, or keep only the 183 tasks with source proofs. Which way should we go?

Let’s preserve the current 233 tasks by using the existing dataset/manifest as the default selection, without adding a new hard-coded allow-list.

Thanks a lot! LGTM.

PR #90, opened 2026-07-30
Remove duplicate proof-completion tasks

Qian-Cheng-nju:
## Summary

- remove 23 byte-identical proof-completion target copies while retaining canonical tasks
- add one post-generation duplicate gate that auto-removes the approved copied-module families listed in JSON and rejects unknown duplicate groups before deleting anything
- update dataset counts to 68 examples and 683 tasks


Qian-Cheng-nju:
@sukanya1426 Could you review that? Thanks a lot!

sukanya1426:
@Qian-Cheng-nju Sure..

source/tlaplus_examples_SpecifyingSystems_Liveness/InternalMemory.tla
```diff
@@ -13,7 +13,7 @@ TypeInvariant ==
   /\ buf \in [Proc -> MReq \cup Val \cup {NoVal}]
 
 Req(p) == /\ ctl[p] = "rdy" 
-          /\ \E req \in  MReq : 
+          /\ \E req \in  MReq :
```
sukanya1426, line 16: I think this commit can be dropped. The two InternalMemory_proof_* tasks (InvInductive and TypeCorrect) were already byte-identical across all four directories, so the gate catches them without changing source/.

It also leaves benchmark/ out of sync. The sources were edited, but the generated models weren't regenerated, so the next generate --shared-model will produce a diff that's missing from this PR.

source/ also appears to be a direct copy of upstream (110 of the 184 specs still contain trailing whitespace), so editing it loses that provenance. This applies equally to the Composing and RealTime copies changed in the same commit.

Could you either drop this commit or regenerate benchmark/ after making the change?

src/dataset/proof_completion/generate.py
```diff
@@ -1120,6 +1122,65 @@ def _run_sany_gate(directory):
...
+            "  - " + "\n    ".join(os.path.relpath(path, directory) for path in group) for group in unknown
+        )
+        raise RuntimeError(f"duplicate task gate found {len(unknown)} unapproved exact-byte group(s):\n{detail}")
+
+    removed = []
+    for keeper, group in approved:
+        for path in group:
+            if path == keeper:
+                continue
+            os.remove(path)
```
sukanya1426, line 1229: The gate removes task files, but it doesn't clean up dependency modules that are no longer used. After this PR, benchmark/proof-completion/tlaplus_examples_SpecifyingSystems_{Composing,Liveness,RealTime}/ contain 10 model files between them, but no remaining tasks.

Nothing breaks, get_benchmark_files only picks up directories with a top-level goal, so the runner ignores them. They're just dead files now, and the tree still contains three example directories that docs/DATASET.md no longer lists.

Could the gate also prune dependency files that no surviving task needs, and remove directories once they're empty?

src/dataset/proof_completion/duplicate_task_families.json
```diff
@@ -0,0 +1,35 @@
...
+    "target_prefix": "HourClock_proof_",
+    "canonical": "tlaplus_examples_SpecifyingSystems_HourClock",
+    "copies": [
+      "tlaplus_examples_SpecifyingSystems_Composing",
+      "tlaplus_examples_SpecifyingSystems_Liveness",
+      "tlaplus_examples_SpecifyingSystems_RealTime"
+    ]
+  },
+  {
+    "target_prefix": "InternalMemory_proof_",
```
sukanya1426, line 27: Just a heads-up since this overlaps with #92, not a change request. Running this against the #92 tree exposed two duplicate groups that aren't listed here:

Consensus/Consensus_Invariance.tla + tlaplus_examples_{Paxos,byzpaxos}/
Consensus/Voting_QuorumNonEmpty.tla + tlaplus_examples_Paxos/

That means whichever PR merges second will need to update this list. If this one lands first, I'll make the change in #92.

src/dataset/proof_completion/generate.py
```diff
@@ -1168,9 +1230,10 @@ def main():
         if count:
             print(f"  -> {count} benchmarks")
 
+    duplicates = _run_duplicate_gate(BENCHMARK_DIR)
     dropped = _run_sany_gate(BENCHMARK_DIR)
-    detail = f" ({total} generated, {dropped} dropped as degenerate)" if dropped else ""
-    print(f"\nTotal benchmarks generated: {total - dropped}{detail}")
+    detail = f" ({total} generated, {len(duplicates)} duplicates and {dropped} degenerate tasks dropped)"
```
sukanya1426: detail is unconditional now, so even a run with nothing to report prints … (0 duplicates and 0 degenerate tasks dropped). The previous behavior only showed this when there was something to report, which seems a bit cleaner. Same on line 1212 in the --shared-model branch.

sukanya1426 (changes requested):

Qian-Cheng-nju:
> Just a heads-up since this overlaps with https://github.com/specula-org/tlaps-bench/pull/92

@sukanya1426 Thanks for pointing this out. Since these duplicate groups are introduced by #92, I think the allowlist should be updated in that PR.

I have fixed other comments, thanks a lot!

Qian-Cheng-nju:
@sukanya1426 Once you think it looks good, please merge.

sukanya1426 (approved):
LGTM!

sukanya1426:
Hi @Qian-Cheng-nju , the changes are looking good to me, but I think I don't have write access to the repository, so I'm unable to merge the PR.

PR #91, opened 2026-07-30
Refine benchmark cost and time accounting

Qian-Cheng-nju:
## Summary

- record formal agent wall time and equivalent public API cost for every benchmark backend
- use native agent telemetry where available and maintained public pricing otherwise
- keep infra and quota attempts in separate diagnostic accounting instead of formal benchmark totals
- report missing or incomplete cost as unavailable rather than zero


src/evaluator/runner.py
```diff
@@ -1344,6 +1614,7 @@ def _run_backend_container(
         install_script=backend.install_script,
         credential_mounts=backend.get_credential_mounts(),
         keep_container=item.keep_container,
+        agent_start_marker=(f"__TLAPS_BENCH_AGENT_START_{uuid.uuid4().hex}__" if _supports_cost_time(backend) else ""),
```
munimthahmid, line 1617: Nit: `--timeout` is described as an agent timeout, but it also includes CLI installation and firewall setup. Could we call it a backend timeout instead?

munimthahmid (approved):
@Qian-Cheng-nju  Reviewed the  implementation thoroughly. Also I ran  Codex, Claude Code, and one shot OpenAI calls on a small proof, including complete and timeout paths . Wall time, token/request accounting, fail-closed behavior, and headline equivalent costs were reported correctly. 

I found no major issue. Approving.

Qian-Cheng-nju:
Thanks a lot! @munimthahmid Merging.

PR #92, opened 2026-07-30
Add layered proof-completion task generation

sukanya1426:
## What changed

- Split each proof-completion task into three files:
  1. `<base>Model.tla` for the model, assumptions, spec, and fairness.
  2. `<task>Scaffold.tla` for the provided definitions and earlier lemmas.
  3. `<task>.tla` for the target theorem and editable proof region.
  Only `<task>.tla` is submitted, and only its proof region can be edited.
- Added `manifest.json` so every task carries its exact context. This reuses the existing marker and manifest logic from proof-from-scratch instead of duplicating it. Every generated task is validated before the manifest is written.
- Regenerated the full dataset with `--layered` (**473 tasks**).
- The corpus exposed two required fixes:
  1. The scaffold now stops at the target theorem to avoid later `VARIABLES` declarations shadowing theorem variables.
  2. Module-level `USE/HIDE` statements now live in the task file, since they are not inherited through `EXTENDS`.
- Small fixes:
  - `parse_theorems` now handles standalone `PROOF OBVIOUS`.
  - `scripts/dataset_table.py` now counts tasks from the manifest instead of counting scaffold files.

## Scope

This PR covers the generator and regenerated dataset only, the evaluator changes are in #89. Now that every task has a manifest, the legacy fallback and prompts in #89 are no longer needed for this suite. It also adds `source/OpenAddressing/OpenAddressing_proof.tla`. The task was already in the dataset, but its source module was missing from `source/`, so regeneration would have removed it. This restores the missing source.

## Testing

- Generated **473/473** tasks. The 13 extra candidates are the same degenerate cases the triviality gate has always filtered out.
- All **473** tasks parse under standalone SANY using only their manifest context. No scaffold is shared between tasks.
- **427/473** reference proofs pass. Every failure matches the pre-regeneration dataset, so there are no regressions.
- The evaluator correctly rejects edits outside the proof region (spec, scaffold, theorem, imports, markers, context files, module-level declarations, and `PROOF OMITTED`). The reference proof still passes.
- Regeneration is byte-for-byte reproducible. A `--filter` run only regenerates its own group and leaves the rest of the manifest unchanged.
- All tests pass: **1,042 unit** and **16 integration** tests, including **27 new generator tests**.

## Issue

`find_proof_end` still misses theorem lines that end right after the `THEOREM` keyword. This can cause `validate.py` to include the next theorem in the task and report a false "1 missing" for some `Sets_*` and `GraphTheorem_*` tasks. This is a pre existing issue and is unchanged by this PR.


sukanya1426:
@Qian-Cheng-nju , Hi, can you please review this pr?

Qian-Cheng-nju:
I'm reviewing that. @munimthahmid Since you have implemented evaluator, could you also help with reviewing this?

src/dataset/proof_completion/generate.py
```diff
@@ -1106,6 +1148,571 @@ def generate_shared_model_l1(output_root=None):
...
+    source it regenerated and leaves every other stored task untouched.
+    """
+    existing = sm._load_existing_manifest(output_root) if incremental else {}
+    all_bases, processed_bases = scope or ({}, {})
+    superseded = sm.tasks_owned_by(existing, all_bases, processed_bases) if incremental else set(existing)
+
+    dropped = []
+    if run_gates:
+        sm.layered_sany_gate(output_root, manifest, audit_writer)
+        dropped = sm.layered_triviality_gate(output_root, manifest, audit_writer)
```
munimthahmid: I ran full generation and got 474 tasks instead of the existing 473. The extra `Data/SequencesTheorems_AppendDef.tla` task passed unchanged with `PROOF OBVIOUS`. The triviality gate treats a `tlapm` timeout as if the task needs a proof. We should retry or stop generation instead of keeping it.

src/dataset/proof_completion/generate.py
```diff
@@ -1106,6 +1148,571 @@ def generate_shared_model_l1(output_root=None):
...
+    base_module = os.path.splitext(os.path.basename(source_path))[0]
+    planned = _plan_targets(sm, dump, source_lines, base_module, subdir, reference_task_keys, audit_writer)
+    if not planned:
+        return 0
+    targets = [target for target, _module, _key in planned]
+
+    os.makedirs(out_dir, exist_ok=True)
+    written = audit_state.setdefault("written", {})
+
+    model_set, _main_specs = sm.compute_model_set(dump, targets)
```
munimthahmid, line 1524: The shared model is built from all targets, so it moves some definitions and assumptions that appear later in the source ahead of earlier theorems. I found 52 tasks with this scope change. For example, `Voting_QuorumNonEmpty` can now use `Ballot`, even though `Ballot` is defined after it. `BY QuorumAssumption DEF Ballot` fails in the old task with `Ballot not found`, but passes in the layered task. We should preserve each target’s original scope and only include model content available before that theorem.

src/dataset/proof_completion/generate.py
```diff
@@ -1106,6 +1148,571 @@ def generate_shared_model_l1(output_root=None):
...
+    is the one the grader will reach. A partial run replaces only the tasks whose
+    source it regenerated and leaves every other stored task untouched.
+    """
+    existing = sm._load_existing_manifest(output_root) if incremental else {}
+    all_bases, processed_bases = scope or ({}, {})
+    superseded = sm.tasks_owned_by(existing, all_bases, processed_bases) if incremental else set(existing)
+
+    dropped = []
+    if run_gates:
+        sm.layered_sany_gate(output_root, manifest, audit_writer)
```
munimthahmid: Generation still reports success when a source or SANY check fails. I tested an invalid task: the gate reported `1/1 FAILED`, but finalization kept it in `manifest.json`. I also passed a nonexistent source file; generation logged the SANY error, exited `0`, and wrote an empty manifest. We should keep the audit output, but return nonzero and not write the final manifest when generation has errors.

munimthahmid (changes requested):
@sukanya1426 Thanks for the work on this. I found a few issues during testing and left inline comments. Could you please check and fix them?

Qian-Cheng-nju:
@sukanya1426 Could you address the comments? Thanks

sukanya1426:
@munimthahmid Thanks for the review, Fixed the issues, can you please take a look once again?
@Qian-Cheng-nju  and sorry for the delay, the generation runs are long and I had a few false starts at first. :")

sukanya1426:
The dataset is now up to 450 tasks. Each task parses correctly using only the context listed in its manifest, and all 1,206 unit tests are passing. I haven't re-run validate, so I removed the old 427/473 numbers rather than keeping outdated results.

src/dataset/proof_completion/generate.py
Qian-Cheng-nju: Thanks for adding the retry. If the retry also times out, the task is still kept and the manifest is written. Since the retry still runs 16 jobs, this may still be resource contention. Please treat a second timeout as a generation error and leave the existing dataset unchanged.

Qian-Cheng-nju:
@sukanya1426 I found three other blockers. After merging with current main, the manifest still has only 450 entries, so the 210 Ben-Or and 48 Tendermint tasks are silently skipped. Also, a failed generation can delete files still referenced by the old manifest, and a validation failure can leave an invalid manifest on disk because it is overwritten before validation. Please update from current main and regenerate all 708 tasks, then make finalization preserve the existing dataset until all checks and manifest validation have passed.

src/dataset/proof_completion/generate.py
```diff
@@ -1108,6 +1150,766 @@ def generate_shared_model_l1(output_root=None):
...
+            audit_writer.write(f"[audit] generation error: {message}\n")
+        audit_writer.write(f"[audit] generation FAILED with {len(errors)} error(s) — manifest not written\n")
+        print(f"\n❌ generation failed with {len(errors)} error(s); manifest not written:", file=sys.stderr)
+        for message in errors[:20]:
+            print(f"   {message[:200]}", file=sys.stderr)
+        if len(errors) > 20:
+            print(f"   ... and {len(errors) - 20} more (see the audit log)", file=sys.stderr)
+        raise SystemExit(1)
+
+    manifest_path = os.path.join(output_root, MANIFEST_FILENAME)
```
munimthahmid, line 1828: Finalization is still not safe. Generated files are written directly into the existing dataset, `sweep_unreferenced_context` can delete files referenced by the old manifest, and `manifest.json` is overwritten before the new manifest is validated. If generation or validation fails, the previous valid dataset is already broken.

A failed run should leave the existing dataset byte-for-byte unchanged. We should generate into a staging directory, run all gates, sweeping, and manifest validation there, and replace the existing dataset only after everything passes. For `--filter`, staging should start from the current valid dataset so skipped tasks remain unchanged.

src/dataset/proof_completion/generate.py
```diff
@@ -1108,6 +1150,766 @@ def generate_shared_model_l1(output_root=None):
...
+        if not _is_structured_proof(sm, target_thm, source_lines):
+            if reference_task_keys is None or task_key not in reference_task_keys:
+                continue
+            audit_writer.write(
+                f"[audit] {task_key}: source states no reference proof — retained for existing-dataset selection\n"
+            )
+        elif reference_task_keys is not None and task_key not in reference_task_keys:
+            audit_writer.write(
+                f"[audit] {task_key}: source candidate is outside the existing dataset selection — generated anyway\n"
+            )
```
munimthahmid, line 1475: As @Qian-Cheng-nju noted, after merging current `main`, the old 450-task manifest skips the 210 Ben-Or and 48 Tendermint tasks.

I also ran fresh generation with `--skip-gates` and got 803 candidates instead of the reviewed 708. After running the triviality check, 61 tasks that #95 excluded because their reference proofs fail still remain.

So updating and regenerating is not enough. Please preserve the exact reviewed 708-task selection and fail if any task is missing or unexpected.

src/dataset/proof_from_scratch/generate.py
```diff
@@ -1784,21 +1806,55 @@ def layered_triviality_gate(output_root, manifest, audit_writer, jobs=16, timeou
...
+    # "a no-op submission cannot PASS". A shorter retry would prove nothing.
+    retry_timeout = retry_timeout or max(timeout, resolve_timeout(None))
+
     def check(item):
         task_key, entry = item
         with tempfile.TemporaryDirectory(prefix="layered_triv_") as tmp:
             path = _materialize_task(output_root, task_key, entry["context"], tmp)
-            degenerate, detail = check_task(path, tlapm_path, tlapm_lib, timeout)
-            return task_key, degenerate, detail
+            return task_key, check_task(path, tlapm_path, tlapm_lib, timeout)
```
munimthahmid: I checked all 56 tasks that use Community Modules. This check runs on a temporary copy and cannot find `lib/community`, while the evaluator provides that library.

Because of this, `SumSequence_Lemma2a.tla` is kept after a missing-module error even though it passes unchanged in the evaluator. Please use the same Community Modules path as the evaluator and treat missing-module errors as generation failures.

munimthahmid (changes requested):
@sukanya1426 Thanks for the updates. The previous inline issues are fixed. I found a few new blockers after merging current `main` and left inline comments. Could you please check and fix them?

sukanya1426:
@Qian-Cheng-nju Done. Finalization now stages the entire run in a temp directory and only swaps it over the dataset after every gate, the sweep, and manifest validation pass, so a failed or invalid run leaves the existing dataset byte-for-byte unchanged. Regenerating from current main now yields all 708 tasks, with the 210 Ben-Or and 48 Tendermint tasks included.

src/dataset/proof_completion/generate.py
sukanya1426: Fixed. Candidates outside the reviewed selection are now skipped instead of generated, and finalization fails if the run doesn't reproduce the exact 708-task selection. The selection is the union of the manifest and the un-migrated flat Ben-Or/Tendermint tasks, so nothing should be silently dropped.

sukanya1426:
@munimthahmid Thanks again. Solved the mentioned issues ...

src/dataset/proof_completion/generate.py
sukanya1426: @Qian-Cheng-nju, a second timeout now counts as a generation failure. The task is discarded, the run fails, and the current dataset and manifest are left untouched rather than including a task the gate never successfully evaluated.

PR #94, opened 2026-07-31
Add portable TLAPS agent skills

munimthahmid:
## What changed

- Add two portable TLAPS skills: one for validating inductive invariants and one for practical proof hints.
- Remove the duplicated skill guidance from the prompt and docs.
- Copy the skills into the project skill folder used by Codex, Claude Code, Copilot, Cursor, LiteLLM, and Pi.
- Save the exact skills used with each run and reuse them on retries.
- Centralize skill discovery and validation so every supported agentic backend receives the same validated skill catalog.
- LiteLLM has no built-in skill loader, so show its agent the catalog and let it read the matching `SKILL.md` when needed.
- Keep one-shot backends unchanged. They make one model request without tools, so they cannot load a matching `SKILL.md` during the run.

## Testing

- All 1,231 automated tests pass.
- Tested skill discovery with all six supported agentic backends: Codex, Claude Code, Copilot, Cursor, LiteLLM, and Pi.
- Live tests confirmed that Cursor and LiteLLM load the matching `SKILL.md`.
- A live Claude Code agent run selected and loaded the invariant-validation skill without being explicitly prompted to use a skill.


munimthahmid:
@Qian-Cheng-nju can you review it? Thanks!

Qian-Cheng-nju:
@sukanya1426 Could you please review this for me first? We should focus mainly on whether the skill installation is correct. I will review this as well.

src/evaluator/backends/pi.py
```diff
@@ -786,7 +787,7 @@ def build_command(self, workspace: str, result_dir: str) -> list[str]:
             (
                 "prompt=$(cat); "
                 f"cd {shlex.quote(workspace)}; "
-                "pi --mode json --no-session "
+                "pi --mode json --no-session --approve "
```
Qian-Cheng-nju: Could we avoid trusting the whole workspace just to load the skills? `--approve` also enables project-level Pi settings and extensions. Since continuations reuse the same workspace, files created in one round could affect the next one.

Please use `--no-approve --skill .agents/skills` instead, and add a test confirming that the skills load while a project extension does not.

src/evaluator/agent_skills.py
```diff
@@ -0,0 +1,124 @@
...
+@dataclass(frozen=True)
+class AgentSkill:
+    """One validated skill in a project catalog."""
+
+    name: str
+    description: str
+    source_dir: Path
+
+
+class _UniqueKeyScalarLoader(yaml.BaseLoader):
```
Qian-Cheng-nju: Could we keep the original YAML value types here instead of converting every scalar to a string? With `BaseLoader`, values such as `name: true` or `description: 123` pass the later string checks, even though Copilot will skip those skills.

Please reject boolean, numeric, and null values, while still accepting explicitly quoted strings such as `"true"` and `"123"`.

skills/tla-inductive-invariant-validation/SKILL.md
```diff
@@ -0,0 +1,107 @@
...
+
+====
+```
+
+Match every re-declared name exactly so implicit `INSTANCE` substitution binds it. Put each `\* @type:` annotation between the `CONSTANT` or `VARIABLE` keyword and the name it annotates, including every name originally declared in a multi-name block. A comment above the keyword is not accepted and leads to `Expected a type annotation for VARIABLE <name>`.
+
+Define `IndInv` in the wrapper. Fix finite constants in `APTarget.cfg`, for example:
+
+```tla
+CONSTANT N = 5
```
Qian-Cheng-nju, line 69: This configuration example also needs `INIT` and `NEXT`. As written, the three documented Apalache commands stop before checking anything with `Neither INIT and NEXT, nor SPECIFICATION found`.

```suggestion
INIT Init
NEXT Next

CONSTANT N = 5
```

src/evaluator/backends/pi.py
munimthahmid: Good point, you’re right. I added `--approve` because Pi doesn’t automatically discover project skills in an untrusted workspace, but it trusts more project-local resources than needed.

I changed it to `--no-approve --skill .agents/skills`.

src/evaluator/agent_skills.py
munimthahmid: Yeah, you’re right. `BaseLoader` converts every scalar to a string, so the later type checks weren’t effective. I switched the duplicate-key loader to `SafeLoader` so the original YAML value types are preserved.

skills/tla-inductive-invariant-validation/SKILL.md
munimthahmid: You’re right. I reproduced the same error with the pinned Apalache 0.58.3. I added `INIT Init` and `NEXT Next` to the example and verified that all three commands now work.

Qian-Cheng-nju:
LGTM! Thanks!

Issue #93, opened 2026-07-31
Add Ben-Or and Tendermint to Proof Completion


Qian-Cheng-nju:
Add Proof Completion tasks from [`ben-or83`](https://github.com/konnov/apalache-examples/tree/main/ben-or83) and [`tendermint`](https://github.com/konnov/apalache-examples/tree/main/tendermint). 

@sukanya1426

- [ ] Import the proof modules and required local dependencies, recording the upstream commit, license, and provenance.
- [ ] Generate the candidates with the existing Proof Completion flow, and keep the data reproducible.
- [ ] Confirm the selected reference proofs pass with the repository's TLAPM and the generated tasks pass the existing dataset validation, then update the dataset summary.

sukanya1426:
Hi @Qian-Cheng-nju , I think we can close this issue now ... Again I dont have the permission to do so 😬

PR #95, opened 2026-07-31
Add proof-completion tasks from ben-or83 and tendermint

sukanya1426:
## What changed

- Bundle only the modules each proof reaches. Skip MC_* configs, add the source commit, licenses, and citations to NOTICE, and bundle Variants.tla from Apalache so tlapm can resolve it.
- Generated 341 candidate tasks. Dropped 22 where `PROOF OBVIOUS` already verifies.
- Dropped another 61 tasks whose reference proofs don't verify with the pinned `tlapm`. Reproduction steps are in `docs/apalache_examples_unverifiable.json`.
- Added 258 tasks (210 from ben-or83, 48 from tendermint), so that brings the proof-completion dataset to 708 tasks across 70 examples.
- Updated `README.md`, `docs/DATASET.md`, and `scripts/dataset_table.py`.

## Testing

- Validated all 258 tasks with `tlaps-bench validate` using the pinned `tlapm`.
- Passed SANY and integrity checks. Every target byte matches its source, with no proof content beyond the target.
- Regeneration is byte identical and only adds the two new example directories.
- 1,202 tests pass; 3 Docker integration tests are skipped in native CI.

## Notes for review

- 55 of the 103 tendermint tasks are excluded because their reference proofs don't verify with the pinned `tlapm`. The sources remain in `source/`, so the tasks can be restored once those proofs verify.
- ben-or83 contributes 210 tasks (~30% of the proof-completion suite). We should probably decide whether we're okay with one benchmark making up that much of the dataset.

Closes issue #93


sukanya1426:
@Qian-Cheng-nju Can you please review the pr? Thanks!

Qian-Cheng-nju:
I did some edits and now it looks good to me. Thanks!

PR #97, opened 2026-08-04
Record how many tools an agent called

lemmy:
Record how many tools an agent called, and which TLA+ tool each reached for.

Every agentic backend now reports a tally through the parse_run_metadata hook the runner already merges into result.json:

    "tool_calls": {"total": 88, "tlaps": 38, "tlc": 0, "apalache": 0,
                   "other": 50}

The split answers a second question the data could not: whether agents reach for the checkers the container ships. Across 659 recorded sessions here they called tlapm 46022 times, Apalache 35, and TLC twice.

Recognition works on the executable of each command in a pipeline, not a substring of the line, so `grep tlapm notes.txt` is not a prover run; command substitution and shell keywords are boundaries too, because `out=$(tlapm ...)` and `for g in ...; do tlapm ...; done` are how agents actually invoke it. Java-hosted tools resolve by main class, since tla2tools.jar holds TLC, the SANY parser and the PlusCal translator alike.

### DIFF
--- a/src/evaluator/backends/cursor.py
+++ b/src/evaluator/backends/cursor.py
@@ -9,6 +9,7 @@ import subprocess
 from ipaddress import ip_address
 from urllib.parse import urlparse
 
+from evaluator import toolcalls
 from evaluator.usage import UsageSummary
 
 from .agentic import AgenticBackend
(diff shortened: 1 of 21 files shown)


Qian-Cheng-nju:
@munimthahmid I reviewed and added some changes. Could you review that?

src/evaluator/backends/cursor.py
```diff
@@ -54,6 +55,109 @@ def _result_usage(event: dict[str, object]) -> UsageSummary | None:
...
+            result_count += 1
+            continue
+        if result_count and event_type in {"assistant", "system", "tool_call"}:
+            activity_after_result = True
+        if event_type != "tool_call":
+            continue
+        subtype = event.get("subtype")
+        raw_call = event.get("tool_call")
+        if not isinstance(raw_call, dict) or not raw_call or len(raw_call) != 1:
+            lifecycle_warnings.append("Cursor tool call has an invalid tool_call object")
```
munimthahmid: I tested this with the current Cursor CLI. The tool call completed successfully with matching `started` and `completed` events and a clean terminal result, but it was reported as `complete=false` and `is_lower_bound=true`.

```text
total=1
tlaps=1
available=true
complete=false
is_lower_bound=true
warnings=["Cursor tool call has an invalid tool_call object"]
```

The emitted event had this shape:

```json
{
  "type": "tool_call",
  "subtype": "started",
  "call_id": "...",
  "tool_call": {
    "shellToolCall": {
      "args": {
        "command": "tlapm --version"
      }
    },
    "hookAdditionalContexts": [],
    "toolCallId": "...",
    "startedAtMs": "..."
  }
}
```

Current Cursor events include metadata alongside `shellToolCall`, so the `len(raw_call) != 1` check marks valid successful calls as incomplete. We should ignore the metadata when identifying the tool type.

munimthahmid (changes requested):
I reviewed the implementation and tested it against all six agentic backends: Claude Code, Codex, Copilot, Cursor, Pi, and LiteLLM. Found an issue with Cursor . I left an inline comment about that issue. Other than that, the tool-call accounting worked correctly.

src/evaluator/backends/cursor.py
Qian-Cheng-nju: Fixed in `ec205d15`: Cursor now identifies the unique `*ToolCall` field and ignores sibling metadata regardless of order; regression tests added.

PR #102, opened 2026-08-07
Structural-complexity metrics for Proof Completion

sukanya1426:
Analysis only branch , not intended to merge. Sharing the writeup and numbers for discussion
## What we did

We went through all 706 Proof Completion tasks in the current manifest and scored each human reference proof on four structural metrics:

- reference proof steps, with a separate Direct bucket for plain `BY ...` proofs
- max proof depth
- transitive proof dependencies
- reference proof obligations from `tlapm --summary`

Also manually checked a mix of cases like, Direct vs structured, shallow vs deep, low vs high deps, few vs many obligations and compared against Composer 2.5 on the tasks that overlap.

## What we found

On the whole, proofs aren’t huge. The typical structured proof is around 10 steps. There is a long tail though, the biggest one I noticed is 1,433 steps and 2,780 obligations.

The main takeaway is which metrics actually add information:

- Steps and obligations basically move together (ρ ≈ 0.97). If we already rank them by steps, obligations don’t give us a new difficulty type.
- Depth also grows with size, so it’s useful as a shape check, but I think not as a second leaderboard axis on its own.
- Dependencies are the interesting extra signal. They’re only weakly tied to proof length, and they catch things steps miss, for example, a Direct proof that looks tiny but leans on a big lemma stack.

## Files
- `analysis/structural-complexity/ANALYSIS.md`: full writeup
- `analysis/structural-complexity/metrics.csv`: per-task numbers
- `scripts/structural_complexity_pc.py`: recompute script


Qian-Cheng-nju:
Thanks for putting this analysis together. My current preference is to use reference proof steps, with Direct proofs in their own category, as the primary complexity signal. We can keep dependencies, proof depth, and obligations as supporting information. This gives us a simple main measure while preserving the extra context for Core selection.

@Saadmrp1038 @munimthahmid, could you both take a look and let us know whether you agree? If so, we can use this.

munimthahmid:
@Qian-Cheng-nju Sounds good to me. For Proof Completion, reference proof steps are clearly the simplest primary complexity signal.

About Direct proofs, I would have suggested removing them since models like Sol can solve them pretty easily. In the last run, 134 out of 135 passed, and the one failure was a session failure. But the recent addition of Ben-or and Tendermint includes some heavier Direct proofs with many more dependencies, so they will likely be harder. Should we keep the heavier Direct proofs from Ben-or and Tendermint and pick only a few good ones from the other 135, instead of keeping every Direct proof? I also think we should calibrate the complexity after multiple comparable runs based on pass rate, time taken to solve, etc.

Qian-Cheng-nju:
> I would have suggested removing them 

Much agree. Removed here: #104

Qian-Cheng-nju:
Closing this as an analysis-only PR. The actionable Core/task-list changes have landed. Thanks for putting the metrics and writeup together.

PR #103, opened 2026-08-07
Add specification-equal scoring

### LINKED ISSUE #101: Add specification-equal scoring for Proof Completion
@munimthahmid Could you define stable specification identities and add an specification-equal score for proof completion(IF you have time, you can also add for proof from scratch in advance)?

The current task-micro pass rate gives every proof task equal weight, so specifications with more extracted tasks have more influence. The primary score should instead give each specification the same total weight while preserving partial credit:

```text
spec_score = passed selected tasks / applicable selected tasks
overall_score = mean(spec_score)
```

Please:

- add a stable, auditable `task_id` to `spec_id` mapping for every task in the current versioned manifest, documenting the grouping rule and flagging ambiguous cases;
- implement the formula above as an scorer, with focused tests for partial credit, missing mappings, and non-applicable results;
- retain task-micro pass rate and all-leaves-complete as secondary diagnostics;
- compare task-micro and specification-macro scores on the same cohort, reporting any meaningful changes in model gaps or ranking.

For this issue, I would like to keep the scope to specification identity and scoring. Core-task selection can follow separately.

munimthahmid:
## What changed

- Add a stable `spec_id` to every task in the versioned manifests: 706 proof-completion tasks and 250 proof-from-scratch tasks.
- Group tasks by the exact source-relative `.tla` path. This keeps files with the same basename separate and treats different source modules in the same example as different specifications.
- Do not count `SKIP`, infrastructure failures, or tasks removed from the current manifest. Report each separately instead.
- Keep the old task-micro score as the primary score when `--scoring equal` is used.

## Scores

- **Specification-macro (primary):** Calculate the pass rate within each specification, then average those rates so every specification has equal weight.
- **Task-micro:** Divide the number of passed tasks by the number of applicable tasks, so every task has equal weight.
- **All-leaves-complete:** Count specifications where every applicable selected task passed.

## Existing result check

I ran the new scorer on the existing Codex GPT-5.5 proof-completion results:

- **Specification-macro:** 95.1% across 95 specifications.
- **Task-micro:** 407/445 tasks passed (91.5%).
- **All-leaves-complete:** 85/95 specifications passed every selected task (89.5%).
- 38 old results were not present in the current manifest, so they were reported as non-applicable and excluded.

## Testing

- Added tests for partial credit, missing mappings, non-applicable results, `SKIP`, and infrastructure failures.
- Regenerated both benchmark suites and verified that all 706 proof-completion tasks and all 250 proof-from-scratch tasks match the versioned outputs.
- Ran Codex GPT-5.5 on `GCD_GCD1` and `GCD_GCD3`. Both passed. Because both tasks come from the same specification, the scores were 100% specification-macro, 2/2 task-micro, and 1/1 all-leaves-complete.

Closes #101.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,10 +85,12 @@ uv run tlaps-bench run --jobs 10 --timeout 7200
 uv run tlaps-bench run --mode proof-from-scratch --jobs 10
 ```
 
-Each run writes `results.json` and `summary.md` (with the headline pass rate);
-`uv run tlaps-bench score` (re)computes and compares pass rates. Use `--resume`
-with a fixed `--output-dir` to skip tasks already recorded as PASS, and
-`--force-build` to rebuild the image after changing source.
+Each run writes `results.json` and `summary.md` (with specification-macro as the
+primary score and task-micro as a diagnostic); `uv run tlaps-bench score`
+(re)computes and compares scores. See the [scoring documentation](docs/USAGE.md#tlaps-bench-score)
+for the grouping rule and formulas. Use `--resume` with a fixed `--output-dir` to
+skip tasks already recorded as PASS, and `--force-build` to rebuild the image
+after changing source.
 
 Choosing an agent (`--backend` / `--model`) and its credentials, the full CLI
 reference, and native (`--no-container`) setup are covered in the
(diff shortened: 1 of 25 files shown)


munimthahmid:
@Qian-Cheng-nju can you look into it? Thanks

Qian-Cheng-nju:
LGTM, thanks a lot!

PR #107, opened 2026-08-09
Make specification pass rate primary

munimthahmid:
Based on a recent Slack conversation and the resulting decision, benchmark scoring now uses the strict specification pass rate as the primary score.


Qian-Cheng-nju:
LGTM! Thanks!

PR #110, opened 2026-08-11
Fix UpdateStateValue precondition

### LINKED ISSUE #109: TPaxosWithProof_UpdateStateValue is false when p = q
## Context

GPT-5.6 Sol agentic xhigh failed this task:

`tlaplus_examples_TencentPaxos/TPaxosWithProof_UpdateStateValue`

The task has no valid proof because the target theorem is false as written.

## Why the theorem is false

`UpdateState` contains these two updates:

```tla
![q] = new_state_qq,
![p] = new_state_qp
```

The theorem allows `p = q`. In that case, both updates target the same entry, so the second update overwrites the first.

Consider a one-participant instance with `B \in Value` and this initial system state:

```tla
msgs = {}
state[q][q] = [maxBal |-> -1, maxVBal |-> -1, maxVVal |-> None]
```

Use these parameters for `UpdateState`:

```tla
p = q
pp = [maxBal |-> 20, maxVBal |-> 15, maxVVal |-> B]
```

These values satisfy the theorem's assumptions, including `Inv`, `pp \in State`, and `pp.maxBal >= pp.maxVBal`.

With these values, `UpdateState` computes:

```tla
new_state_qq = [maxBal |-> 20, maxVBal |-> -1, maxVVal |-> None]
new_state_qp = [maxBal |-> 20, maxVBal |-> 15, maxVVal |-> B]
```

Therefore, the final value is:

```tla
state'[q][q] = [maxBal |-> 20, maxVBal |-> 15, maxVVal |-> B]
```

The theorem claims that either:

1. `maxVBal` and `maxVVal` remain unchanged, but they change from `-1` and `None` to `15` and `B`; or
2. `pp.maxVBal = pp.maxBal` and `state'[q][q].maxBal = pp.maxVBal`, which would require `15 = 20` and `20 = 15`.

Neither alternative holds. All the theorem's assumptions are true while its conclusion is false.

Qian-Cheng-nju:
## Summary
- require `p # q` in `UpdateStateValue`
- discharge the new precondition from `MsgInv` in `UpdateStateMsgInv`
- regenerate the affected Proof Completion scaffolds without changing the Full (706) or Core (293) cohorts

Fixes #109

### DIFF
--- a/benchmark/proof-completion/tlaplus_examples_TencentPaxos/TPaxosWithProof_AcceptMsgInvScaffold.tla
+++ b/benchmark/proof-completion/tlaplus_examples_TencentPaxos/TPaxosWithProof_AcceptMsgInvScaffold.tla
@@ -239,7 +239,8 @@ LEMMA UpdateStateViewValue ==
 PROOF OMITTED
 
 LEMMA UpdateStateValue ==
-          ASSUME NEW q \in Participant, NEW p \in Participant, NEW pp \in State, pp.maxBal >= pp.maxVBal,
+          ASSUME NEW q \in Participant, NEW p \in Participant, NEW pp \in State, p # q,
+                     pp.maxBal >= pp.maxVBal,
                      UpdateState(q, p, pp), Inv
             PROVE \/ /\ state'[q][q].maxVBal = state[q][q].maxVBal
                      /\ state'[q][q].maxVVal = state[q][q].maxVVal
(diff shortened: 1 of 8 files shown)


munimthahmid (approved):

munimthahmid:
The fix is correct, all TLAPS obligations now pass, and the generated tasks are consistent. LGTM!

PR #118, opened 2026-08-14
Repair Proof-from-Scratch dataset integrity

munimthahmid:
### Specification fixes

- Fixed OpenAddressing so its special `empty` value cannot be mistaken for a fingerprint.
- Required at least two fingerprints, preventing division by zero in the fingerprint calculation.
- Regenerated the OpenAddressing files for both benchmark modes and added a bounded TLC model covering all five properties.
- Fixed etcd Raft's `PendingMessages` definition. The old fold mixed up a message with the result collected so far and could read `.msource` from an empty sequence.

### Dataset cleanup

- Removed five duplicate Cantor tasks, reducing Proof-from-Scratch from 250 to 245 tasks.
- Kept `Cantor1_cantor` and `Cantor5_cantor` as the two distinct proof problems.

<details>
<summary>Why the Cantor tasks were duplicates</summary>

- `Cantor1_cantor` through `Cantor4_cantor` contain the same proof problem after ignoring generated module names and formatting.
- `Cantor5_cantor` through `Cantor7_cantor` form a second identical group.
- Their read-only definition files contain no additional assumptions or definitions that would make the tasks different. The Cantor source files and Proof-Completion tasks are unchanged.

</details>

- Generated read-only files now stop at the outer `====`, while nested modules are preserved.
- Generation now fails if read-only context contains trailing content, theorems, or proof text.
- Regeneration removed unintended trailing content from 12 context files.

### Result impact

The old Proof-from-Scratch results should not be reused for the latest 245-task set.

<details>
<summary>Why the old results need to be rerun</summary>

- The task set changed from 250 to 245 tasks.
- The OpenAddressing and etcd inputs changed, along with 12 generated context files.
- During the audit, I found 15 previous PASS results across nine tasks that used exposed theorems whose proofs were marked `OMITTED`.

A clean run of all 245 tasks is needed so every reported model is evaluated on the same updated inputs.

</details>

## Testing

- Regenerated all 245 selected tasks in the pinned container with no missing or unexpected tasks.
- None of the tasks passed with the placeholder proof.
- Full and filtered regeneration reproduced the committed files.
- Standalone SANY passed all 245 task bundles.
- The integrity check passed all 690 generated TLA+ files.
- A bounded TLC model checked all five OpenAddressing properties without errors: 903 generated states, 491 distinct states, depth 18.
- Focused generator, contract, integrity, and evaluator tests: 198 passed.
- `uv run ruff check .`, `uv run ruff format --check .`, and `git diff --check` passed.


munimthahmid:
@Qian-Cheng-nju @sukanya1426, can you review it? Thanks!

sukanya1426:
Everything seems to be working well as described.

Qian-Cheng-nju:
LGTM. Thanks a lot.

> During the audit, I found 15 previous PASS results across nine tasks that used exposed theorems whose proofs were marked OMITTED.

That's alright. They're still valid tasks.

PR #119, opened 2026-08-14
fix quantify ZooKeeper LocalPrimaryOrder over every pair

sukanya1426:
## What was the issue :
Local primary order is supposed to mean, if a primary broadcasts a before b, a follower that delivers b must already have delivered a. The formula used `\E txn1, txn2` , “there exist two transactions”. In TLA+ those two names can be the same transaction, and a transaction is always equal to itself, so once there are a couple of broadcasts the formula is already true. It never actually checks order. The task was not asking for the property in the comment / Zab paper.

## How we solved : 
I switched that to `\A` for all pairs in:
- `source/ZooKeeper/Zab.tla`
- `source/ZooKeeper_LowLevel/ZkV3_7_0.tla`
- the two matching proof-from-scratch Defs files

and added a test so we don’t silently go back to `\E`.
I checked this with TLC on a tiny example using the same comparison operators as Zab: broadcasts a then b, follower log has delivered b and never a (`<<x, b>>`). The old `\E` formula is true in that state, the new `\A` formula is false. That’s what we want, the paper property fails, and the old encoding does not notice. I did not model check the full ZooKeeper spec (`Spec => []LocalPrimaryOrder`). That spec is too open ended for TLC as we ship it.

If we’d rather not ship a stronger theorem without a full-spec check, we can drop these two tasks instead of this fix.

## Testing:
- TLC on the small `\E` vs `\A` example
- SANY on both staged LocalPrimaryOrder task bundles
- Integrity check on the proof-from-scratch suite
- `pytest tests/dataset/test_zookeeper_local_primary_order.py`

### DIFF
--- a/source/ZooKeeper/Zab.tla
+++ b/source/ZooKeeper/Zab.tla
@@ -1075,9 +1075,8 @@ LocalPrimaryOrder == LET p_set(i, e) == {p \in proposalMsgsLog: /\ p.source = i
                              \/ /\ ~TxnEqual(txn1, txn2)
                                 /\ LET TxnPre  == IF ZxidCompare(txn1.zxid, txn2.zxid) THEN txn2 ELSE txn1
                                        TxnNext == IF ZxidCompare(txn1.zxid, txn2.zxid) THEN txn1 ELSE txn2
-                                   IN \A j \in Server: /\ lastCommitted[j].index >= 2
-                                                       /\ \E idx \in 1..lastCommitted[j].index: 
-                                                            TxnEqual(history[j][idx], TxnNext)
+                                   IN \A j \in Server: (\E idx \in 1..lastCommitted[j].index: 
+                                                            TxnEqual(history[j][idx], TxnNext))
                                         => \E idx2 \in 1..lastCommitted[j].index: 
                                             /\ TxnEqual(history[j][idx2], TxnNext)
                                             /\ idx2 > 1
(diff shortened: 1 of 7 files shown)


sukanya1426:
@munimthahmid can you please take a look? Thanks!

source/ZooKeeper/Zab.tla
```diff
@@ -1070,7 +1070,7 @@ LocalPrimaryOrder == LET p_set(i, e) == {p \in proposalMsgsLog: /\ p.source = i
                      IN \A i \in Server: \A e \in 1..currentEpoch[i]:
                          \/ Cardinality(txn_set(i, e)) < 2
                          \/ /\ Cardinality(txn_set(i, e)) >= 2
-                            /\ \E txn1, txn2 \in txn_set(i, e):
+                            /\ \A txn1, txn2 \in txn_set(i, e):
```
munimthahmid, line 1073: The `\A` change correctly fixes the existential/self-pair tautology, but I think the repair is incomplete. The implication still requires `lastCommitted[j].index >= 2`, so a follower that has committed only `<<b>>` after broadcasts `a` then `b` is skipped and `LocalPrimaryOrder` evaluates to `TRUE`.

I reproduced this with a small TLC test that directly evaluates the formula with broadcasts `a` then `b` and follower history `<<b>>`. The proposed formula returns `TRUE`, even though the property stated directly above `LocalPrimaryOrder` requires a follower delivering `b` to have delivered `a` first. This does not establish that the state is reachable under the full `Spec`, but it shows that the predicate itself remains weaker than the property it is intended to encode.

I think we should remove the `>= 2` conjunct in both definitions and generated Defs files. The existing `\E idx ... TxnNext` already determines whether `b` was delivered, and `idx2 > 1` already requires an earlier entry. The regression test should also include this semantic case, since its current substring assertions do not catch it.

munimthahmid (changes requested):

source/ZooKeeper/Zab.tla
sukanya1426: Agreed, I have addressed it now.

PR #123, opened 2026-08-15
fix: do not skip singleton log ZooKeeper TotalOrder

sukanya1426:
Follow up to #119. 

What was wrong :

Total order is supposed to mean, if some follower delivers a before b, then any process that delivers b must also deliver a, and deliver a before b. The formula required `committed2 >= 2` on that other process. So if one log was `<<a, b>>` and another was only `<<b>>`, the implication never fired and the property held. TLC on that pair: old formula true, new formula false.
The witness still needs two entries (`committed1 >= 2`). We kept that. `GlobalPrimaryOrder`'s `>= 2` is also left alone, that one really needs two deliveries on the same log.

What changed :

Dropped `committed2 >= 2` in both ZooKeeper sources and both Proof-from-Scratch Defs modules.

Testing :

- Checked the << b >> case with TLC, the old formula said true, the new one says false. `<<a, b>>` / empty / singleton witness still succeed.
- There is a lock test so committed2 >= 2 cannot sneak back in. The LocalPrimaryOrder tests still pass.
- SANY on both staged TotalOrder task bundles
- Integrity / spec identity / ruff 

Note : Did not perform TLC `Spec => []TotalOrder` on the full ZooKeeper spec.

### DIFF
--- a/benchmark/proof-from-scratch/ZooKeeper/ZabModel.tla
+++ b/benchmark/proof-from-scratch/ZooKeeper/ZabModel.tla
@@ -32,6 +32,17 @@ ACK       == "ACK"
 COMMIT    == "COMMIT"
 Quorums == {Q \in SUBSET Server: Cardinality(Q)*2 > Cardinality(Server)}
 
+BootstrapZxid == <<0, 1>>
+BootstrapTxn == [ zxid   |-> BootstrapZxid,
+                  value  |-> 0,
+                  ackSid |-> Server,
+                  epoch  |-> 0 ]
+BootstrapProposalMsgs ==
+    { [ source |-> s,
+        epoch  |-> 0,
+        zxid   |-> BootstrapZxid,
+        data   |-> 0 ] : s \in Server }
+
 VARIABLES state,          
           zabState,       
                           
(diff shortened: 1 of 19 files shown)


sukanya1426:
@munimthahmid Hi! Can you please take a look ? Thankss.....

munimthahmid:
The change looks correct, but after thinking about it more, I don’t think we should make this change in the benchmark. The same concern applies to PR #119 

Neither original task was shown to be invalid. Both PRs strengthen property definitions inherited from the upstream ZooKeeper specification and hence change the meaning of the theorems. I think it is better for the benchmark to remain faithful to upstream and, if needed, report these proposed changes there first. If upstream accepts them, we can update our copy from the new upstream version.

What do you think, @Qian-Cheng-nju?

Qian-Cheng-nju:
Thanks! That's interesting. I will talk to Lingzhi Ouyang. Let's keep this open for now.

Qian-Cheng-nju:
Confirmed with the upstream developer Lingzhi Ouyang: this is an initialization issue, not a problem with the original TotalOrder formula. I reworked the PR accordingly. It now restores the original TotalOrder guard and initializes both ZooKeeper models with a common committed bootstrap entry representing the already-established session and target key.

Qian-Cheng-nju:
FYI: https://github.com/Disalg-ICS-NJU/zookeeper-tla-spec/pull/2

PR #124, opened 2026-08-15
Restore etcd Raft domain assumptions

munimthahmid:
Found one more invalid task (hopefully the last one :) ). It was one of the tasks GPT-5.5 timed out on: `etcd_raft_MoreThanOneLeader`. I checked the other timed-out tasks again and did not find another invalid one.

Basically, the old flat benchmark had `FiniteServers` and `NilNotServer`, but our layered etcd source and generated model were missing them. Without `NilNotServer`, `Nil` can be a server ID, allowing the same server to vote twice and two leaders to be elected in the same term.

Counterexample: with `Server = InitServer = {1, 2, 3}` and `Nil = 1`, TLC reaches a state where servers 1 and 2 are both leaders in term 1.

Fix: restored `FiniteServers` and `NilNotServer` while keeping `InitServerSubset`, and added regression coverage for the missing assumptions.

Testing: the original counterexample is now rejected before TLC computes initial states

### DIFF
--- a/benchmark/proof-from-scratch/etcd_raft/etcd_raftModel.tla
+++ b/benchmark/proof-from-scratch/etcd_raft/etcd_raftModel.tla
@@ -514,6 +514,10 @@ Spec == Init /\ [][Next]_vars
 
 ASSUME InitServerSubset == InitServer \subseteq Server
 
+ASSUME FiniteServers == IsFiniteSet(InitServer)
+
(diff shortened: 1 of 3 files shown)


munimthahmid:
CC: @Qian-Cheng-nju

Qian-Cheng-nju:
LGTM. Thanks!

PR #127, opened 2026-08-19
Expose proof libraries to proof-from-scratch tasks

### LINKED ISSUE #126: proof-from-scratch tasks can be unprovable because the TLAPS proof libraries are out of scope
A proof-from-scratch task inherits its `EXTENDS` clause from the source spec, and the agent has no legal way to add one. When the source spec does not import `TLAPS`, `NaturalsInduction`, `FiniteSetTheorems` or `WellFoundedInduction`, the agent cannot cite `NatInduction`, `FS_Induction` or `WFInduction`. For theorems whose proof needs e.g. a well-founded measure, that plausibly makes the task unprovable within the time budget.

## Why the agent cannot work around it

Three independent enforcement points, none of which is individually wrong:

1. `src/dataset/proof_from_scratch/generate.py` copies the source spec's `EXTENDS` verbatim  into the generated model module and adds nothing. `FlashWithMutex.tla:13` declares `EXTENDS Naturals, FiniteSets`, so `FlashWithMutexModel.tla:3` does too, and the `Defs` and task modules just extend down the chain.
2. `src/evaluator/prompts/proof-from-scratch.txt:45` forbids adding `EXTENDS` or `INSTANCE`, and `src/tlacheck/rules/helper_region.py` classifies an `INSTANCE` in the helper region as `Severity.CHEATING`.
3. The existing `EXTENDS` line sits outside the editable regions, so rewriting it trips `SCAFFOLD_MODIFIED` in `src/common/check_proof.py`.

Note that the task contract itself has no objection to these modules:`RESOLVABLE_MODULES = STDLIB_MODULES | COMMUNITY_MODULES` in `src/common/tla_modules.py` whitelists the entire standard library for reference. The gap is only that the generator never puts them in scope.

Backend pragmas are *not* part of the problem: `PTL`, `SMTT`, `ExpandENABLED` and friends are plain definitions, so an agent can redefine them in the helper region, and they do — Opus 5 proved all 7 German tasks that way, and the current Flash run drives `expandenabled`, `isabelle`, `zenon`, `smt3` and `ls4` through locally defined operators. Library *theorems* have no such workaround.

## Why `--allow-no-proof` makes it worse

For a task generated from a theorem with an upstream proof, the spec's imports are known to suffice for at least one proof. Tasks generated with `--allow-no-proof` come from theorems upstream leaves `PROOF OMITTED`, so nothing establishes that the inherited import surface is adequate for *any* proof. `FlashWithMutex_DirProgressCorrect` is that case: 16 of the 17 Flash theorems have been proved by a model under the current imports, and the one that needs a measure induction over a shrinking set has resisted four multi-hour attempts by frontier models. In the current attempt the agent identifies the missing induction principle explicitly and spends its budget trying to derive one via `Isa`.

munimthahmid:
## What changed

In proof-from-scratch tasks, agents could not use standard induction theorems when the source spec did not import their libraries. Task imports are read-only, so agents could not add the missing libraries themselves.

- Add TLAPS and the natural-number, finite-set, and well-founded induction libraries to every generated task.
- Put the library declarations in the read-only part of the task, so agents can use them without changing imports.
- Regenerate all 245 tasks and update the prompts and tests with the available theorem names.

## Testing

- Regenerated and checked all 245 proof-from-scratch tasks.
- An end-to-end test confirms that agents can use `NatInduction`, `FS_Induction`, and `WFInduction`, while agent-added imports remain rejected.

Closes #126.

### DIFF
--- a/benchmark/proof-from-scratch/Allocator/Allocator_InitMutex.tla
+++ b/benchmark/proof-from-scratch/Allocator/Allocator_InitMutex.tla
@@ -1,5 +1,11 @@
 ---- MODULE Allocator_InitMutex ----
 EXTENDS Allocator_InitMutexDefs
+
+LOCAL INSTANCE TLAPS
+LOCAL NatInductionLib == INSTANCE NaturalsInduction
+LOCAL FiniteSetTheoremsLib == INSTANCE FiniteSetTheorems
+LOCAL WellFoundedInductionLib == INSTANCE WellFoundedInduction
+
 \* BEGIN AGENT HELPERS
 \* END AGENT HELPERS
 THEOREM InitMutex == Init => Mutex
--- a/benchmark/proof-from-scratch/Allocator/Allocator_InitTypeInvariant.tla
+++ b/benchmark/proof-from-scratch/Allocator/Allocator_InitTypeInvariant.tla
@@ -1,5 +1,11 @@
 ---- MODULE Allocator_InitTypeInvariant ----
 EXTENDS Allocator_InitTypeInvariantDefs
+
+LOCAL INSTANCE TLAPS
+LOCAL NatInductionLib == INSTANCE NaturalsInduction
+LOCAL FiniteSetTheoremsLib == INSTANCE FiniteSetTheorems
+LOCAL WellFoundedInductionLib == INSTANCE WellFoundedInduction
+
 \* BEGIN AGENT HELPERS
 \* END AGENT HELPERS
 THEOREM InitTypeInvariant == Init => TypeInvariant
--- a/benchmark/proof-from-scratch/Allocator/Allocator_NextMutex.tla
+++ b/benchmark/proof-from-scratch/Allocator/Allocator_NextMutex.tla
@@ -1,5 +1,11 @@
 ---- MODULE Allocator_NextMutex ----
 EXTENDS Allocator_NextMutexDefs
+
+LOCAL INSTANCE TLAPS
+LOCAL NatInductionLib == INSTANCE NaturalsInduction
+LOCAL FiniteSetTheoremsLib == INSTANCE FiniteSetTheorems
+LOCAL WellFoundedInductionLib == INSTANCE WellFoundedInduction
+
 \* BEGIN AGENT HELPERS
 \* END AGENT HELPERS
 THEOREM NextMutex == TypeInvariant /\ Mutex /\ Next => Mutex'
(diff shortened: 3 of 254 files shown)


munimthahmid:
CC: @Qian-Cheng-nju

Qian-Cheng-nju:
Thanks! LGTM

lemmy:
I think this needs to address three concerns:

1. **Symbol collisions:** library modules may export names that clash with the original module.
2. **Incomplete and evolving library set:** the proposed list is already incomplete—TLAPS includes other theorem libraries such as [FiniteSetTheorems.tla](https://github.com/tlaplus/tlapm/blob/main/library/FiniteSetTheorems.tla), and [CommunityModules](https://github.com/tlaplus/CommunityModules/tree/master/modules) contains additional ones. Since these libraries may continue to grow, any hard-coded list will also become stale.
3. **Agent performance:** extending irrelevant modules, e.g. `WellFoundedInduction` when well-foundedness is unnecessary, may add distracting definitions and proof techniques and make proof search harder.

There is some tension here with cheating prevention, but ideally the agent should discover available library modules and their symbols dynamically, then extend only what it needs. For example, SANY's XML Exporter can already enumerate the modules available on the library path.

Qian-Cheng-nju:
Sure, thanks for your reminder. We will rework on that soon!

munimthahmid:
Your concerns are absolutely valid. I actually had similar concerns in mind while designing this, but I still went with the current architecture because letting the agent import arbitrary modules could increase the risk of cheating. For example, it could import a `_proofs` module containing a `PROOF OMITTED` theorem and use that admitted theorem as a fact.

But your points are totally valid. This architecture is not very scalable as we add more specs and libraries, and it may somewhat distract the agent with libraries it does not need, although the prompt does explain which libraries are available.

Letting the agent discover the available libraries and import only what it needs seems like the better direction. I will investigate more into this.

Qian-Cheng-nju:
@lemmy I’m sorry about this issue. This was my fault. I reviewed the specifications early on but failed to revisit our assumptions as more tasks without human proofs were added. Before claiming release, I should have ensured both that successes reflect genuine model capability rather than cheating, and that failures reflect model limitations rather than unreasonable benchmark restrictions. 

My proposed solution #128(still WIP) is to let agents import official libraries while the grader verifies that those libraries come from official sources and have not been modified. We will warn users when newer upstream versions are available, but will not force upgrades so benchmark runs remain reproducible and comparable. Thank you for identifying this issue. I will carefully review the results next time. What do you think of this solution? CC: @munimthahmid

lemmy:
> My proposed solution #128(still WIP) is to let agents import official libraries while the grader verifies that those libraries come from official sources and have not been modified. We will warn users when newer upstream versions are available, but will not force upgrades so benchmark runs remain reproducible and comparable. Thank you for identifying this issue. I will carefully review the results next time. What do you think of this solution?

Makes total sense.

Qian-Cheng-nju:
@lemmy FYI, the current proof-library setup relies on the agent inspecting the available libraries and deciding what to import, so one-shot cannot use it. We have therefore disabled one-shot for proof-from-scratch for now.

PR #128, opened 2026-08-20
Allow proof-from-scratch agents to import official proof libraries

Qian-Cheng-nju:
## Summary

- pin exact TLAPM and CommunityModules source snapshots, warn when upstream moves, and generate a frozen official-module catalog at run start
- allow named `LOCAL <alias> == INSTANCE <module>` declarations inside `AGENT HELPERS`, while rejecting non-official modules, `WITH`, unnamed/out-of-region imports, and workspace shadowing
- remove the four fixed proof-library imports from all 245 proof-from-scratch tasks and record corpus/library digests to prevent incompatible resume


Qian-Cheng-nju:
@munimthahmid Could you review this? Thanks a lot!

src/common/proof_libraries.py
```diff
@@ -0,0 +1,335 @@
...
+class OfficialLibraryCatalog:
+    """Official module names and bytes frozen before an agent starts."""
+
+    sources: dict[str, dict[str, str]]
+    modules: dict[str, dict[str, str]]
+    digest: str
+
+    @property
+    def allowed_modules(self) -> frozenset[str]:
+        return frozenset(self.modules)
```
munimthahmid, line 154: `FiniteSetTheorems_proofs` is the only `_proofs` module in the pinned libraries where I found an explicit `PROOF OMITTED`: `TwoExpLemma`. With the current catalog, an agent can import that module, cite the omitted lemma, and still get a PASS. Are we intentionally trusting this omitted lemma because the file comes from the pinned TLAPM repository? Also, do we need to make `_proofs` modules importable at all? I thought the goal was to let agents use the public library modules, while `_proofs` files are their internal implementations.

src/common/check_proof.py
```diff
@@ -572,6 +583,30 @@ def canonical_replay_required(explicit):
...
+    if benchmark_dir:
+        canonical_names = {
+            os.path.splitext(os.path.basename(path))[0] for path in glob.glob(os.path.join(benchmark_dir, "*.tla"))
+        }
+    issues = []
+    for path in glob.glob(os.path.join(os.path.dirname(filepath), "*.tla")):
+        if os.path.abspath(path) == os.path.abspath(filepath):
+            continue
+        module_name = os.path.splitext(os.path.basename(path))[0]
+        if module_name in official_modules and module_name not in canonical_names:
```
munimthahmid, line 599: This allows a same-name module when it comes from the task's read-only context. But TLAPS searches the task context first. For example, if the task contains `Functions.tla` and the agent imports `FunctionTheorems`, TLAPS can load the task's `Functions.tla` instead of the pinned one, while the checker still accepts it. None of the current 245 tasks have this name conflict, so this does not affect the current dataset. I reproduced it with a test task and the checker returned PASS. I am raising it because future tasks could hit the same problem. Should we verify the exact files loaded for the imported module and all the modules it depends on?

src/evaluator/runner.py
```diff
@@ -2401,6 +2525,15 @@ def main():
             2, f"{parser.prog}: error: no benchmarks found for mode {mode.name!r} under {mode.benchmark_dir()}\n"
         )
 
+    proof_library_catalog = None
+    run_identity = None
+    if mode.name == "proof-from-scratch":
+        try:
+            proof_library_catalog = scan_official_libraries()
```
munimthahmid: In container mode, this scans `lib/tlapm` and `lib/community` on the host before the Docker image is built or checked. Those directories do not exist in a fresh checkout, and native setup is documented as optional. I reproduced this and the run failed before Docker started. Could container mode build or check the image first, then create the catalog from the libraries inside the image?

src/evaluator/modes/proof_from_scratch.py
```diff
@@ -71,3 +72,8 @@ def get_dependencies(self, benchmark_path: str) -> list[str]:
                 f"{benchmark_path}"
             )
         return [str(path) for path in boundary.context_paths]
+
+    def one_shot_prompt_values(self, benchmark_path: str) -> dict[str, str]:
+        path = Path(benchmark_path).parent / CATALOG_FILENAME
+        catalog = load_frozen_catalog(path) if path.is_file() else scan_installed_libraries()
+        return {"proof_library_modules": ", ".join(sorted(catalog.allowed_modules))}
```
munimthahmid: The one-shot prompt receives only module names. Since a one-shot agent cannot inspect the library files, it may know that `WellFoundedInduction` exists but not that the theorem is named `WFInduction`. I think we should include the module-to-theorem list in the one-shot prompt. Although this means including the whole list in every one-shot task's prompt, I don't see a better way because one-shot agents cannot inspect the libraries.

src/evaluator/runner.py
```diff
@@ -1051,6 +1075,88 @@ def _write_task_list_record(output_dir: str, mode_name: str, task_ids: list[str]
...
+    )
+    return f"{value}-dirty" if dirty.returncode == 0 and dirty.stdout.strip() else value
+
+
+def _proof_from_scratch_run_identity(mode: Mode, catalog: OfficialLibraryCatalog) -> dict[str, object]:
+    return {
+        "schema_version": 1,
+        "mode": mode.name,
+        "benchmark_revision": _benchmark_revision(),
+        "execution_source_digest": _image_source_fingerprint(),
```
munimthahmid, line 1200: The library source commits are pinned, but the TLAPM executable still comes from the rolling `1.6.0-pre` release. This means two machines could use different TLAPM binaries while recording the same run identity. Should we pin the executable checksum, or include and verify the actual TLAPM and SANY hashes in this run identity?

munimthahmid (changes requested):

src/common/proof_libraries.py
Qian-Cheng-nju: Good catch. These are implementation modules, not part of the interface we want to expose. The catalog now skips `*_proofs` modules, so `FiniteSetTheorems_proofs` is no longer importable.

src/common/check_proof.py
Qian-Cheng-nju: You're right that this could affect a future task. But I think if one occurs, the agent can avoid that import and copy the proof directly.

src/evaluator/runner.py
Qian-Cheng-nju: Fixed. Container mode now builds or checks the image first and reads the frozen catalog from inside it. A fresh checkout no longer needs host copies of `lib/tlapm` or `lib/community`.

src/evaluator/modes/proof_from_scratch.py
Qian-Cheng-nju: Good point. For now, proof-from-scratch is limited to backends that can inspect the workspace instead of putting the full theorem catalog into every one-shot prompt.

src/evaluator/runner.py
Qian-Cheng-nju: Fixed. TLAPM and SANY are now pinned by SHA-256 in a shared toolchain lock. Native setup and Docker verify the hashes, and the verified toolchain digest is included in the run identity.

src/common/check_proof.py
munimthahmid: Ok, sounds good

munimthahmid:
Looks good to me!

PR #131, opened 2026-08-21
Benchmark/vendor examples specs

lemmy:
Vendors six `tlaplus/Examples` specifications as proof-from-scratch tasks: 245 → 277.

| group | tasks | goals |
| --- | --- | --- |
| `braf` | 14 | invariants, per-action step simulation, `RAF!Spec` refinement, `ENABLED` |
| `bosco` | 5 | `TypeOK`, agreement lemmas |
| `btree` | 5 | structural invariants |
| `Disruptor` | 4 | `TypeOk`, `NoDataRaces` (MPMC and SPMC) |
| `dag-consensus` | 3 | Sailfish `TypeOK`, `Agreement`, `Liveness` |
| `NanoBlockchain` | 1 | upstream `Safety` theorem |

Every goal is an invariant or property the upstream TLC configuration checks. Theorems and, where the constants were unconstrained, assumptions were added; each assumption commit states what breaks without it.

## Refactorings

Committed separately from the verbatim vendoring, one commit each:

- **braf**: three modules split into one file per module — tlapm resolves `EXTENDS` by file name.
- **btree, Nano, Digraph, BlockDag**: recursive operators rewritten as recursive functions — one `RECURSIVE` declaration anywhere in the module chain makes tlapm abort with `Failure("Expr.Anon: Recursive")`, whatever the goal. Justified by TLC exploring the same state space before and after; counts are in the commit messages.

## Generator fixes (prerequisites)

- A multi-line `EXTENDS` whose keyword stands alone lost only its keyword.
- A `RECURSIVE` declaration survived the stripping of its definition.

Regenerating the whole suite is byte-identical for all pre-existing tasks; `tests/dataset` passes.

## Review note

Sailfish's `Agreement` and `Liveness` are the only goals whose truth at arbitrary constants rests on assumptions added here — the Byzantine quorum intersection properties `TLCSailfish1` states in prose. TLC verifies them for both harnesses' constants.


Qian-Cheng-nju:
@lemmy I may be missing an intended constraint, but I can reproduce two counterexamples at `e5a66ca` when `Writers` and `Readers` overlap, which the current `ASSUME`s allow. With `Writers = Readers = {actor}`, `Size = 2`, and `MaxPublished = 4`:

- MPMC violates `NoDataRaces` after `BeginWrite, EndRead, BeginWrite, EndRead, BeginRead`: slot 0 contains `actor` in both its reader and writer sets.
- SPMC violates `TypeOk` after a cross-role `EndWrite` advances `published` without writing the next slot; a later read appends `NULL`, giving `consumed = <<0, 0, NULL>>`, which is not in `Seq(Nat)`.

Are `Writers` and `Readers` intended to be disjoint logical actor IDs? If so, I think the theorem context needs a named assumption such as `WritersReadersDisjoint == Writers \cap Readers = {}`. If overlap is intentional, the shared `pc` appears to need separate writer/reader roles. Please let me know if I have overlooked a constraint or modeled the intended identities incorrectly.

lemmy:
> @lemmy I may be missing an intended constraint, but I can reproduce two counterexamples at `e5a66ca` when `Writers` and `Readers` overlap, which the current `ASSUME`s allow. With `Writers = Readers = {actor}`, `Size = 2`, and `MaxPublished = 4`:
> 
> * MPMC violates `NoDataRaces` after `BeginWrite, EndRead, BeginWrite, EndRead, BeginRead`: slot 0 contains `actor` in both its reader and writer sets.
> * SPMC violates `TypeOk` after a cross-role `EndWrite` advances `published` without writing the next slot; a later read appends `NULL`, giving `consumed = <<0, 0, NULL>>`, which is not in `Seq(Nat)`.
> 
> Are `Writers` and `Readers` intended to be disjoint logical actor IDs? If so, I think the theorem context needs a named assumption such as `WritersReadersDisjoint == Writers \cap Readers = {}`. If overlap is intentional, the shared `pc` appears to need separate writer/reader roles. Please let me know if I have overlooked a constraint or modeled the intended identities incorrectly.

Let's get confirmation from the spec's original author: https://github.com/tlaplus/Examples/pull/232

lemmy:
https://github.com/tlaplus/Examples/pull/230
https://github.com/tlaplus/Examples/pull/231
https://github.com/tlaplus/Examples/pull/232
https://github.com/tlaplus/Examples/pull/233

munimthahmid:
I found three issues and verified all three with TLC by reproducing counterexamples.

- The upstream SPMC model uses `Writers = {w}`, but the theorem context allows multiple writers. With two writers, TLC violates both `NoDataRaces` and `TypeOk`.
- The upstream B-tree config gives the control states distinct model values, but that constraint is missing from the theorem context. With only `SPLIT_ROOT_LEAF` and `SPLIT_ROOT_INNER` aliased, TLC violates both `InnersMustHaveLast` and `KeyOrderPreserved` while every other upstream value remains unchanged.
- The Sailfish assumptions require quorum intersection, but they do not require the set of correct nodes to be a quorum. With an upward-closed quorum definition satisfying every current assumption, TLC reaches a state where a correct node is at round `r + 2` but the correct leader is absent from its log, violating `Liveness`. The upstream configurations rule out this case through their concrete quorum definitions.

sukanya1426:
I found one more issue in btree. MaxOccupancyPositive allows MaxOccupancy = 1, and with Vals = {vx}, MaxKey = 3, and MaxNode = 8, TLC finds a KeyOrderPreserved violation in 24 steps. Node 3 points to node 1 through both key slots, while node 3 holds key 1 and node 1 holds key 2, breaking the ordering. Node 2 also remains reachable but is marked free. MaxOccupancy = 2 is clean at both MaxNode = 8 and 12, so this is a branching-factor issue, not a node-budget one. Upstream's 9126be8 already requires MaxOccupancy >= 2, the vendored copy predates that fix.

Everything else looked good, I checked several valid bosco (N,T,F) combinations, including T > F and multi-million-state runs, Sailfish's GST, all 14 braf goals against the upstream config, and the generated tasks' assumption chain.

Two smaller assumption gaps :

- Nano's KeyPair: The vendored version allows multiple private keys to share a public key, unlike upstream's surjective constraint. This increases the state space from 3,003 to 61,055 states, but Safety still holds.
- Disruptor's NULL: NULL is unconstrained, and NullNotAValue isn't enforced through module instantiation. Even NULL = 0 runs clean because the protocol writes before reading.

Neither breaks a goal, but both weaken the assumptions compared with upstream. Since I only ran TLC, not tlapm, I can't say whether the goals are actually provable under these weaker assumptions.

lemmy:
@munimthahmid and @sukanya1426, may I suggest reproducing the issue and reporting your findings upstream, i.e., in the `tlaplus/Examples` repository? That would allow the original authors to review the findings and sign off on any proposed changes.

The PRs for [Braf](https://github.com/tlaplus/Examples/pull/230) and [Sailfish](https://github.com/tlaplus/Examples/pull/233) are still open, so they can be amended. The others may require new PRs.

Also, I just [pushed](https://github.com/tlaplus/Examples/commit/c757b12b7a4de6fa39523db541133fece9ced416) the obvious strengthening of SPMC's assumption.

Qian-Cheng-nju:
@lemmy Since tlaplus/Examples#230 has received no review or feedback since August 24, I went ahead and completed the downstream update rather than leave this PR blocked indefinitely.

I applied #230’s proposed Braf assumptions as written, synchronized the merged tlaplus/Examples#233 Sailfish changes,including CorrectNodesFormQuorum.

I also rechecked Braf and TLCSailfish1 with TLC. They completed with 128,581/3,316 and 314,144/109,604 total/distinct states respectively, with no errors. The previous Sailfish counterexample configuration is now rejected by CorrectNodesFormQuorum.

Unless you would prefer to wait for upstream review of #230, I think this PR is ready to move out of draft and merge.

lemmy:
@Qian-Cheng-nju Thanks, I went ahead and merged https://github.com/tlaplus/Examples/pull/230

Issue #132, opened 2026-08-21
proof-from-scratch should emit one task per module, not one per theorem


lemmy:
The generator emits one module per theorem, deleting every other theorem of the source module — statement and proof — and pruning definitions to the closure of the target's statement. A task cannot cite a sibling result, and where the target's statement does not mention it, cannot even name the predicate.

The emitted task is then not the one the author solved. `FindHighest.tla` states four theorems, all four tasks, and the last one is two steps upstream:

```tla
THEOREM IsCorrect == Spec => []Correctness
<1>1. TypeOK /\ InductiveInvariant /\ DoneIndexValue => Correctness
  BY DEF TypeOK, InductiveInvariant, DoneIndexValue, Correctness
<1>. QED
  BY <1>1, TypeInvariantHolds, InductiveInvariantHolds, DoneIndexValueThm, PTL
```

`FindHighest_IsCorrect` drops the three cited theorems and the definitions `TypeOK`, `InductiveInvariant`, `DoneIndexValue`; its entire context is `vars`, `Init`, `lb`, `Terminating`, `Next`, `Spec`, `Correctness`. Discharging it means reinventing the inductive invariant, and that is reported under the same name as the author's two lines. The duplication is the smaller half: 82 of 277 tasks have an upstream proof citing a deleted sibling, 172 come from a module contributing two or more tasks, and all three tasks of `AsyncTerminationDetection_proof.tla` cite `TypeCorrect`, so it is proved three times, in three containers, uncounted.

Fix: one task per source module. Keep every theorem, `PROOF OMITTED` on each, and let the model choose the order and cite what it has already proved — how the proofs were written in the first place. A module scores k/n theorems. It also removes the
incidental context: `braf`'s `Thm_Safety` today sees `Inv2`, only because four actions conjoin it, but not `TypeOK`, `Inv1`, `Inv3`, `Inv4`, `Inv5`.red at c563ae5.

Qian-Cheng-nju:
Thanks @lemmy, I completely agree with your diagnosis and am working on it.

I see two options:

1. One task per source module, as you proposed. This is simpler and avoids duplicated work. Individual tasks will take longer, though total runtime should decrease. Based on existing per-theorem runtimes, the longest module task could exceed 10 hours.
2. Keep one task per theorem, but run them incrementally in dependency order and carry forward earlier verified results. This keeps tasks shorter but is more complex.

I lean toward the first option. Would you prefer the other approach, or something else?

[...]

PR #133, opened 2026-08-22
Add proof-from-scratch module task contract

Qian-Cheng-nju:
## Summary

- replace the inactive theorem-batch foundation with a minimal one-task-per-module contract
- identify multiple proof regions with existing theorem task IDs while keeping model, definitions, and statements immutable
- define dependency-closed trusted proof units without changing the current corpus or evaluator

### DIFF
--- a/src/common/proof_from_scratch_module.py
+++ b/src/common/proof_from_scratch_module.py
@@ -91,6 +91,7 @@ class ModuleTaskRegions:
 
         editable = [self.helpers if helpers is None else helpers]
         editable.extend(replacements.get(proof.task_id, proof.text) for proof in self.proofs)
+        editable = [body if not body or body.endswith(("\n", "\r")) else body + "\n" for body in editable]
         parts: list[str] = []
         for fixed, body in zip(self.fixed_segments[:-1], editable, strict=True):
             parts.extend((fixed, body))
--- a/src/common/proof_from_scratch_module.py
+++ b/src/common/proof_from_scratch_module.py
@@ -1,0 +1,372 @@
+"""Shared contract for future module-level proof-from-scratch tasks.
+
+The current corpus remains one physical task per target theorem.  A future
+corpus revision will group those same target IDs into one editable proof module
+per source specification.  This module fixes only the small boundary shared by
+the generator and evaluator: strict metadata, identified proof regions, fixed
+scaffold extraction, and dependency-closed trust.  It deliberately performs no
+SANY or TLAPM work.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import re
+from collections.abc import Iterable, Mapping
+from dataclasses import dataclass
+from pathlib import Path, PurePosixPath
+from typing import Any
+
+from common.task_contract import BEGIN_AGENT_HELPERS, BEGIN_AGENT_PROOF, END_AGENT_HELPERS, END_AGENT_PROOF
+
+MODULE_TASK_FORMAT_VERSION = 1
+
+BEGIN_AGENT_PROOF_PREFIX = BEGIN_AGENT_PROOF + " "
+END_AGENT_PROOF_PREFIX = END_AGENT_PROOF + " "
+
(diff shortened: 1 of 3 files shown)


Qian-Cheng-nju:
@munimthahmid @sukanya1426 You could review this as well while you are developing.

src/common/proof_from_scratch_module.py
```diff
@@ -0,0 +1,372 @@
...
+        proofs: Mapping[str, str] | None = None,
+    ) -> str:
+        """Rebuild the task while replacing only identified editable bodies."""
+
+        replacements = proofs or {}
+        unknown = set(replacements) - set(self.proof_unit_ids)
+        if unknown:
+            raise ModuleTaskContractError(f"replacement proofs contain unknown task IDs: {sorted(unknown)}")
+
+        editable = [self.helpers if helpers is None else helpers]
```
sukanya1426, line 92: render() inserts bodies as-is, so a body without a trailing newline gets the closing marker attached to its last line:

```tla
BY DEF Spec, Init, Next, Inv\* END AGENT PROOF W/W_A.tla
```
I checked the impact, SANY accepts it and tlapm still proves all 6/6 obligations, but `parse_module_task_regions` later fails with `must contain exactly one END AGENT PROOF …`. So there's no false-PASS risk, but it can cause a confusing false negative.

Could we normalize this by adding `\n` to non-empty bodies? The empty case needs the guard because the helper region ships empty, and appending there inserts a blank line and breaks `render() == original`.

#136 is the intended caller, and `"\n".join(...)` or `.strip()` can produce this case. The existing test only uses newline-terminated bodies, so this isn't covered.

sukanya1426 (changes requested):

src/dataset/proof_from_scratch/design.md
```diff
@@ -243,6 +243,32 @@ scratch. (Exception: when the target property *is* an invariant, as in
...
 
+## Module-task foundation (Issue #132; not yet active)
+
+The checked-in theorem tasks and their canonical contexts remain the current
+execution contract. A future corpus revision will group the existing manifest
+targets by `spec_id`, producing one task per source module while retaining the
+current task IDs as the scored proof-unit IDs.
+
+Each module task keeps only the definitions in the union of its target
+statements' dependency closures. Original lemmas, non-target theorems, proof
```
munimthahmid, line 254: I might be missing something, but doesn't this still leave part of the problem from #132? For example, `NextMutex` in `Allocator.tla` uses `RequestMutexBis`, `AllocateMutex`, and `ReturnMutex`, but this contract removes them because they are not existing targets. When I tried the original proof with that context, SANY reported all three as unknown operators. Should these supporting theorems stay in the module even if they are not scored?

src/common/proof_from_scratch_module.py
Qian-Cheng-nju: Fixed. Thanks!

src/dataset/proof_from_scratch/design.md
Qian-Cheng-nju: I see proof from scratch as requiring the agent to invent the proof structure from scratch. #132 addresses a different problem: work the agent has already completed should not be discarded. Therefore, target theorems that the agent proves should remain available to later target proofs.

The supporting theorems in your example are not benchmark targets, so I do not think we should provide their statements to the agent. The agent should rediscover them if needed, and once it proves them in the shared helper region, it can reuse them throughout the module.

I may be misunderstanding your concern. Is this the distinction you had in mind?

munimthahmid: Yes, that’s the distinction I was missing. I agree the current contract is correct.

munimthahmid:
I tested this contract against all 109 generated module tasks and 245 proof units. Everything looks good to me now.

Qian-Cheng-nju:
@sukanya1426 Could I merge this?

sukanya1426:
@sukanya1426 Yeah LGTM...

PR #135, opened 2026-08-24
Make SANY an unconditional grading gate

### LINKED ISSUE #134: SANY is not an unconditional gate — invalid specs can pass on tlapm alone
## Problem

`tlapm`'s parser is more permissive than `tla2sany`, and the gap is not cosmetic — it can make `FALSE` provable. Per [tlaplus/tlapm#285 (comment)](https://github.com/tlaplus/tlapm/pull/285#issuecomment-5383137174):

```tla
---- MODULE Foo -----
THEOREM Eq    == [a |-> 1, a |-> 2] = [a |-> 1, a |-> 3] OBVIOUS
THEOREM False == ASSUME NEW r, r = [a |-> 1, a |-> 2] PROVE FALSE OBVIOUS
=====
```

`tlapm` accepts the duplicate record field and reports "All 2 obligations proved": the record axioms are instantiated per field entry, so the duplicated name gets two values. SANY rejects the same constructor outright ("Non-unique fields in constructor"). Upstream, [tlaplus/tlapm#213](https://github.com/tlaplus/tlapm/issues/213) tracks replacing the tlapm parser with SANY; until that lands, SANY is the only thing standing between this benchmark and a graded proof of `FALSE`.

So SANY must gate every graded solution. Today it is enforced only conditionally.

## Where the gate is missing

1. **An unrunnable SANY counts as a passed SANY** — `src/common/check_proof.py:1413`. `check_sany_valid` returns `valid`, `invalid`, or `unavailable`; only `invalid` fails the solution. When SANY cannot run at all (missing Java, `run.sh`, or `tla2tools.jar`, or a 120s timeout) the result is `unavailable`, which emits a warning and lets grading continue to `tlapm --strict` — so the solution can reach PASS unparsed. The hard stop at `:1414` covers only `canonical_replay_required` modes (`src/evaluator/modes/proof_from_scratch.py:24`, `src/evaluator/modes/proof_completion.py:69`), leaving unmarked proof-completion tasks exposed.

2. **A real rejection can be mistaken for an unrunnable SANY** — `src/common/check_proof.py:1016` classifies the outcome by grepping the error text for five phrases; anything unmatched becomes `unavailable` and is therefore graded as if SANY had approved the module. Note that "Non-unique fields in constructor" matches none of the five. The string it inspects is also truncated to 400 characters at `src/tlacore/sany/dump.py:54`, which can cut off the very diagnostics the match depends on.

3. **The same failure quietly weakens tamper detection** — `src/tlacheck/context.py:135` sets `sany_ok` from an error-swallowing helper, and `src/tlacheck/engine.py:42` then drops `statement_modified`, `extra_axiom`, `smuggled_module`, and `helper_region`. One absent toolchain removes both the parse gate and the checks that would catch a weakened target theorem.

4. **Neither case is visible afterwards** — `src/evaluator/runner.py:2364` derives `sany_valid` from the absence of a stdout marker, so a run where SANY never executed is recorded as `true`, indistinguishable from a clean parse. Existing results can't be audited for (1).

5. **`validate_benchmarks.py` never calls SANY** — `src/common/validate.py` runs `tlapm --strict` only (`:197`, `:233`), so an invalid reference proof can enter the benchmark.

## Proposed fix

A grader that cannot run its own parse gate has no verdict to give: make `unavailable` an `ERROR` (exit 3) in every mode rather than a warning. Decide `invalid` from the dumper's exit code (`src/dataset/sany-dump/DumpSemantics.java:51` already exits 3 precisely when `errorLevel > 0`) instead of matching keywords, and stop truncating those diagnostics. Have `tlacheck` distinguish "SANY rejected the module" from "SANY did not run". Record the three-way status in `results.json`. Add the gate to `validate.py`.

## Acceptance criteria

- The `MODULE Foo` module above fails in every mode, rather than passing on `tlapm`'s say-so. Worth adding as a regression fixture.
- A solution `tlapm` accepts but SANY rejects fails in every mode.
- A run with the SANY toolchain removed errors instead of passing.
- `results.json` states whether SANY ran.

## Out of scope

`--keep-verifying` and `--sany-only` are not gaps: the former continues past an `invalid` result but the verdict still fails (`src/common/check_proof.py:1417`, `:1620`), and the latter is an explicit opt-in fast path.

Qian-Cheng-nju:
## Summary

- classify standalone SANY outcomes as valid, invalid, or unavailable from process evidence instead of diagnostic text
- fail closed across grading, `--sany-only`, semantic checks, and startup preflight while recording `sany_status` and a complete `grading/sany.log`
- keep `--sany-only` submission-local while requiring full checks to use an independent canonical target
- require manifest-backed Proof Completion evaluation and add SANY validation to reference-proof validation

Historical results and leaderboard data are unchanged.

Fixes #134


Qian-Cheng-nju:
@sukanya1426 Could you review that? Thank you!

src/common/check_proof.py
```diff
@@ -979,46 +993,51 @@ def record_check_commit(repo_root, tree, message):
...
-        # A genuine rejection carries SANY's own diagnostics; a bare "produced no
-        # dump" without them is an infrastructure failure (run.sh / Java missing),
-        # not an invalid spec — treat that as unavailable, not invalid.
-        low = msg.lower()
-        if any(k in low for k in ("parse error", "errorlevel", "*** errors", "unknown operator", "multiply-defined")):
-            return "invalid", msg
-        return "unavailable", msg
-    except Exception as e:
-        return "unavailable", f"{type(e).__name__}: {e}"
+    return run_normalized(filepath, dep_dir=dep_dir, timeout=120)
```
sukanya1426, line 1023: run_normalized can still raise OSError during mkdtemp/copy2, and nothing catches it now, main caught these and returned "unavailable". When it does escape it's an uncaught traceback that exits 1 (FAIL), so an infrastructure error gets recorded as a failed proof instead of ERROR(3).

It only triggers on the staging path (module name differing from the filename, or multiple dep dirs), benchmark tasks normally hit the fast path, so this is a latent edge rather than something firing now. validate.py:260 has the same unguarded call and its future.result() isn't wrapped, so catching it inside run_normalized would cover both I guess.

src/common/check_proof.py
```diff
@@ -979,46 +993,51 @@ def record_check_commit(repo_root, tree, message):
...
+
+
+def write_sany_log(run: SanyRun, output_path: str) -> str:
+    """Persist the complete SANY process evidence next to the checker result."""
+
+    log_path = os.path.join(os.path.dirname(os.path.abspath(output_path)), "sany.log")
```
sukanya1426: sany.log is always saved as <output-dir>/sany.log, so when multiple targets share a results directory, each run overwrites the previous log. I confirmed this with two targets, both .result files were there, but only one sany.log remained. Using the result filename (e.g. <stem>.sany.log) would fix it.

src/common/check_proof.py
```diff
@@ -1288,26 +1326,27 @@ def main(*, require_canonical_for_proof_from_scratch: bool = False):
                 require_canonical=True,
             )
         try:
-            status, detail = check_sany_valid(sany_file)
+            sany_run = check_sany_valid(sany_file)
         finally:
             if staged_dir:
                 shutil.rmtree(staged_dir, ignore_errors=True)
-        if status == "valid":
+        if args.output:
```
sukanya1426: output_path already has a default, but --sany-only without --output silently skips writing the log, confirmed there's no sany.log. The full check path always writes one, so both paths should check output_path instead of args.output for consistency.

src/tlacheck/context.py
```diff
@@ -131,8 +134,16 @@ def build_context(
     # + solution.tla); the result dir supplies the submission and any
     # agent-created modules, and overrides on name clashes (later wins).
     dep_dirs = [d for d in (benchmark_dir, solution_dir) if d]
-    solution = try_dump_normalized(sol_path, dep_dirs=dep_dirs)
-    sany_ok = solution is not None
+    if sany_run is None:
+        sany_run = run_normalized(sol_path, dep_dirs=dep_dirs)
```
sukanya1426: Reusing the injected sany_run can cause a mismatch, check_sany_valid parses with dep_dir = dirname(filepath), while build_context resolves dependencies from [benchmark_dir, solution_dir]. I got opposite results on the same file, invalid vs. valid. Both shipped modes currently stage into a temp directory, so this is latent rather than actively breaking, but it would be safer to reuse the run only when the dependency directories match.

src/common/check_proof.py
```diff
@@ -583,6 +585,18 @@ def canonical_replay_required(explicit):
     return explicit or os.environ.get("TLAPS_CANONICAL_REPLAY_REQUIRED") == "1"
 
 
+def require_independent_canonical_target(filepath, benchmark_dir):
+    """Reject a canonical target that aliases the submitted target."""
+
+    canonical_path = os.path.join(benchmark_dir, os.path.basename(filepath))
+    try:
+        aliases_submission = os.path.samefile(filepath, canonical_path)
```
sukanya1426, line 594: samefile raises FileNotFoundError when the canonical target is missing, even though I believe that should mean the files aren't aliases. This can be reached from _run_in_container, where benchmark_dir may come directly from args.benchmark_dir without the existence check. Could we treat a missing canonical file as not aliasing instead of failing with exit 3?

src/common/check_proof.py
```diff
@@ -1236,8 +1261,17 @@ def main(*, require_canonical_for_proof_from_scratch: bool = False):
     marked_task = task_has_proof_region_markers(filepath, benchmark_dir)
     marked_proof_completion = args.mode == "proof-completion" and marked_task
     strict_contract = marked_task or (args.mode == "proof-from-scratch" and args.canonical_replay_required)
-    if marked_task:
+    if marked_task and not args.sany_only:
         args.canonical_replay_required = True
+    if args.sany_only:
```
sukanya1426, line 1283: This line makes the if args.canonical_replay_required, staging block unreachable under --sany-only. I confirmed it by making stage_verification_files fail if called, with --sany-only --canonical-replay-required --benchmark-dir <dir>, it never runs and a module whose dependency exists only in the benchmark directory gets rejected.

Nothing breaks today because agent workspaces copy dependencies locally, but the block is effectively dead. We should either remove it or make --canonical-replay-required actually trigger the staging.

sukanya1426 (changes requested):

Qian-Cheng-nju:
@sukanya1426 Fixed all of them. Could you review again? Thanks a lot!

src/common/check_proof.py
```diff
@@ -1045,14 +1070,20 @@ def parse_strict_status(tlapm_exit, tlapm_output):
...
-    benchmark_dir = args.benchmark_dir or resolve_benchmark_dir(None, filepath, target_name)
+    benchmark_dir = None if args.sany_only else args.benchmark_dir or resolve_benchmark_dir(None, filepath, target_name)
     if require_canonical:
         benchmark_dir = resolve_benchmark_dir(benchmark_dir, filepath, target_name)
     if require_canonical and benchmark_dir is None:
         print("ERROR: canonical replay required, but no canonical benchmark directory is available")
         sys.exit(3)
+    if benchmark_dir is not None:
+        try:
+            require_independent_canonical_target(filepath, benchmark_dir)
```
sukanya1426, line 1089: Thanks, I retested everything else, and it all looks resolved. One more thing I noticed, that is the copy back further down in this function still only copies check.result, so sany.log gets left behind when result_dir is removed. I confirmed this under Docker, both files are created in the container, but only check.result makes it back to the host.

Native mode keeps the log because it already writes to the host directory, so the two paths currently behave differently and container mode is the default. Could we copy sany.log back alongside the result?

Qian-Cheng-nju: Fixed. Thanks! @sukanya1426

sukanya1426 (approved):
LGTM! Thanks @Qian-Cheng-nju

PR #136, opened 2026-08-25
Implement module-level proof-from-scratch evaluation

munimthahmid:
Built on the merged #133 and #137.

## What changed

- Run each proof-from-scratch module in one agent session and score its original theorem targets as separate proof units.
- Grade every target and referenced helper against canonical inputs, count a result only when its local dependencies are trusted, and preserve the latest module artifact and checkpoint across timeout, continuation, interruption, and resume.
- Report proof-unit `k/n` and specification aggregates, and reject theorem-level results when resuming a module run.

## Testing

- `uv run ruff check`, `uv run ruff format --check`, and `uv run pytest tests` (2,019 passed, 2 skipped locally; 2,013 passed, 8 skipped in exact-head CI)
- Generated `FindHighest` as one four-unit module, then ran four fresh `gpt-5.6-sol` max-reasoning attempts with no continuation. All four passed 4/4 and persisted their artifacts, checkpoints, results, and reports.
- Ran four different modules with the same setup. `Allocator` passed 4/4, `SimpleMutex` passed 2/2, `AddTwo` passed 2/2, and `BubbleSort` reached 2/3 before the 30-minute timeout. Its partial result was preserved and graded.
- Regenerated and verified all 109 module tasks, then graded every canonical unresolved module through the evaluator. All 109 returned valid structured results with correct proof-unit mappings and 0 errors.
- GitHub `native` CI passed.

PR #137, opened 2026-08-26
proof-from-scratch: one task per module

sukanya1426:
### The problem :
Isolating theorems often removed sibling results they depended on. For example, `FindHighest`'s `IsCorrect` cites three other theorems, but those disappeared when it was turned into a standalone task. This meant the original proof was effectively being evaluated without the context it relied on. This affected 82 tasks, and many others came from the same module, meaning shared work had to be repeated across multiple tasks.

### The change :

Each source module is now a single task containing all its selected theorems. Every theorem gets its own `PROOF OMITTED` region, and the model can choose the proof order and reuse results it has already proved. Dependencies are shared across the module, but copied dependency files are still pruned to only what the task actually needs. This removes 286 unnecessary definitions across 59 files. The benchmark still contains exactly 245 proof units, now grouped into 109 module tasks.

- BubbleSort fix: Combining statements into one module exposed a TLA+ scoping issue in two `BubbleSort` statements. Some locally bound names conflicted with variables declared later in the source.

The generator safely alpha-renames only those binders. The source remains unchanged, and the generated tasks record every rename for auditing.

### Validation : 
Everything was regenerated and verified from scratch
- 243/245 statements are byte-identical to the original corpus, the other two are the required BubbleSort alpha-renames.
- 109/109 module tasks pass SANY and `tlapm`.
- 245/245 proof units are present with no duplicates.
- `FindHighest` successfully proves all 45 obligations using the author's proofs, including sibling theorem references.
- The two renamed BubbleSort statements prove all 25 obligations.
- No definitions were lost compared to the original tasks.
- The corpus and source files remain unchanged.
- No proof artifacts leak into read-only context.

`--verify` regenerates the expected output from the original corpus and compares the entire generated tree.

The PR adds 65 tests, covering generation, validation, dependency pruning, binder renaming, and rejection cases.

### For #136
the generated suite is under `benchmark/proof-from-scratch-module/`. Each manifest entry includes the task spec, its read-only context, and any `renamed_bindings`.


sukanya1426:
@munimthahmid , Hi, can you take a look here ?

src/dataset/proof_from_scratch/module_plan.py
```diff
@@ -0,0 +1,683 @@
...
+import json
+import os
+import sys
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
+from io import StringIO
+from pathlib import Path
+from types import MappingProxyType
+
+PLAN_FILENAME = "module-plan.json"
```
munimthahmid: I think this is a blocker for the batch work. #137 uses a different plan path, schema, and digest format from #133, so our evaluator in #136 cannot load it. You should rebase this onto #133 and use the shared contract.

src/dataset/proof_from_scratch/batch_inputs.py
```diff
@@ -0,0 +1,852 @@
...
+    )
+    if header_index is None:
+        raise BatchInputError(f"TLA+ file has no module header: {path}")
+
+    extends: list[str] = []
+    body_start = header_index + 1
+    for index in range(header_index + 1, len(lines)):
+        match = _EXTENDS_LINE.match(lines[index].rstrip("\n"))
+        if match:
+            extends = [name.strip() for name in match.group(1).split(",") if name.strip()]
```
munimthahmid: `_split_defs_module()` only reads the first line of `EXTENDS`. Multiline declarations produce invalid overlays, and SANY rejects three viable overlays in the current corpus. The full declaration needs to be preserved.

src/dataset/proof_from_scratch/batch_inputs.py
```diff
@@ -0,0 +1,852 @@
...
+            + (f" and {len(divergent) - 4} more" if len(divergent) > 4 else "")
+        )
+
+    if source_preamble and source_preamble != target_preamble:
+        reasons.append("divergent-declarations: the two tasks declare different constants or variables")
+
+    target_module_names = {name for name, _text in target_modules.values()}
+    unresolved = sorted(
+        name
+        for name in source_extends
```
munimthahmid: I believe `carry_forward_conflicts()` is missing dependencies from `INSTANCE` expressions. It marks five `GermanData` edges as viable even though the targets do not contain `GermanControl`, and SANY rejects them after injection.

src/dataset/proof_from_scratch/batch_inputs.py
```diff
@@ -0,0 +1,852 @@
...
+    A definition the two tasks share is not by itself a conflict.  Both were cut
+    from one source module by one pruning pass, so an identically spelled
+    definition IS the same definition: the overlay simply omits it and lets the
+    target's own copy serve, which keeps the carried theorem meaning what it
+    meant.  A shared name whose bodies differ is a conflict and always will be —
+    the theorem would silently change meaning — and so is anything the overlay
+    cannot omit its way around.
+
+    Checked against the canonical corpus, so a batch knows before it runs which
+    carry-forwards are available.  An agent's helper region can only add fresh
```
munimthahmid: I think checking only the canonical definitions is not enough. An accepted helper can add a name that collides with a later task. I reproduced this with `InductiveInvariant == TRUE`, so helper declarations need to be checked against the target before injection.

src/dataset/proof_from_scratch/batch_inputs.py
```diff
@@ -0,0 +1,852 @@
...
+    A PASS means tlapm verified these lemmas, so a later task may use them as
+    facts.  Their proofs are dropped rather than forwarded: the fact is what
+    carries, and shipping the text would hand the next agent a worked example
+    its own budget did not pay for.
+    """
+
+    match = _PROOF_BODY.search(helpers)
+    if match is None:
+        return helpers
+    return helpers[: match.start()] + f"{match.group('indent')}PROOF OMITTED\n"
```
munimthahmid: `_omit_proofs()` returns after the first helper proof and drops everything after it. With two accepted helpers, the second one never reaches the overlay. It should omit each proof separately and keep all helper declarations.

src/dataset/proof_from_scratch/batch_inputs.py
```diff
@@ -0,0 +1,852 @@
...
+        )
+    if document.get("manifest_digest") != manifest_digest(manifest):
+        problems.append("batch inputs were built from a different corpus")
+    expected_plan = _digest_text(json.dumps(_plan_identity(plan), sort_keys=True, ensure_ascii=False))
+    if document.get("plan_digest") != expected_plan:
+        problems.append("batch inputs were built from a different module plan")
+    if problems:
+        return tuple(problems)
+
+    for batch in document.get("batches", ()):
```
munimthahmid: The verifier seems too trusting here. It reports no problem when `batch.json` is missing or corrupted, when a complete manifest has no batches, or when source task bytes have changed. It should derive the expected tree from the current plan and corpus.

munimthahmid (changes requested):

sukanya1426:
@munimthahmid  fixed them, can you take a look again? thanks...

src/dataset/proof_from_scratch/batch_inputs.py
```diff
@@ -0,0 +1,1009 @@
...
+            kept.append(body.rstrip("\n"))
+            continue
+        kept.append(body[: proof.start()].rstrip("\n") + f"\n{proof.group('indent')}PROOF OMITTED")
+    return "\n".join(kept) + "\n" if kept else ""
+
+
+def _declared_helper_names(helpers: str) -> set[str]:
+    """Names an accepted helper region introduces at the top level."""
+
+    return set(_DEFINITION.findall(helpers)) | set(_NAMED_ASSERTION.findall(helpers))
```
munimthahmid: I think `_declared_helper_names()` is still incomplete. The unindented helper is caught now, but an indented `InductiveInvariant == TRUE` is missed. Injection succeeds and SANY then rejects the target because the operator is already defined. You should collect helper names from the parsed module instead of relying on a column-sensitive regex.

src/dataset/proof_from_scratch/batch_inputs.py
```diff
@@ -0,0 +1,1009 @@
...
+    if first is None:
+        return helpers
+    # Anything before the first declaration is kept verbatim: a RECURSIVE line
+    # belongs to the definition that follows it, and dropping it would leave a
+    # recursive operator undeclared.
+    kept: list[str] = []
+    leading = helpers[: first.start()].strip("\n")
+    if leading:
+        kept.append(leading)
+    for declaration in _HELPER_DECLARATION.finditer(helpers):
```
munimthahmid: `_omit_proofs()` still depends on the helper formatting. With an indented `Fresh == TRUE` between two proved lemmas, the accepted input parses, but `Fresh` disappears from the overlay and SANY rejects the second lemma. Same-line `BY` and `OBVIOUS` proofs are also copied instead of omitted.

src/dataset/proof_from_scratch/batch_inputs.py
```diff
@@ -0,0 +1,1009 @@
...
+        batch = recorded_by_id[spec_id]
+        batch_dir = _batch_dir(spec_id)
+
+        batch_path = output_root / PurePosixPath(batch_dir) / BATCH_FILENAME
+        try:
+            declared = json.loads(batch_path.read_text(encoding="utf-8"))
+        except (OSError, UnicodeError, json.JSONDecodeError) as exc:
+            problems.append(f"{batch_dir}/{BATCH_FILENAME} is missing or unreadable: {exc}")
+            declared = None
+        if type(declared) is dict:
```
munimthahmid: I think `verify_batch_inputs()` is still trusting the generated metadata. I changed `depends_on`, `verified_inputs`, step indexes, roots, and file records, and every case still returned clean. A local `Naturals.tla` also passed verification even though it shadowed the standard module and made SANY fail. It should compare the exact expected documents and reject unexpected files.

munimthahmid (changes requested):

munimthahmid:
Hi @sukanya1426, when you get a chance, could you please address the remaining comments on #137? My work on #136 depends on those fixes, so I’m currently blocked on completing the integration. Thank you so much!

sukanya1426:
@munimthahmid I have fixed the issues, do you mind taking a look again? thanks

munimthahmid:
Let's hold until Qian decides on issue #132, which will require a meaningful change to the generator's structure.

Qian-Cheng-nju:
I have finalized the one-task-per-module direction in #132 and replaced the old batch foundation in #133. Please rewrite this PR on top of #133 at `0d3708f5`. The previous overlay and carry-forward design is no longer needed.

### Dataset and generation

@sukanya1426

- [ ] Use the current proof-from-scratch manifest as the only target-selection source and group its 245 task IDs by `spec_id`.
- [ ] Generate one module task per `spec_id`, using the existing layered `Model + Defs + Task` structure.
- [ ] Use the `spec_id` as the module task ID and retain the existing theorem task IDs as ordered proof-unit IDs.
- [ ] Keep the union of definitions reachable from the selected theorem statements.
- [ ] Remove original lemmas, non-target theorems, proof bodies, comments, and proof-only definitions.
- [ ] Add one module-level helper region and one identified `PROOF OMITTED` region per proof unit, using the contract from #133.
- [ ] Produce a deterministic, versioned module-task manifest without modifying the current 245-task corpus.
- [ ] Remove the old isolated theorem inputs, ordered/free strategies, carry-forward analysis, and verified overlay code.

### Validation

- [ ] First validate `FindHighest` as a four-proof-unit canary.
- [ ] Confirm that the generated task is SANY-valid and contains no proof artifacts outside the selected theorem statements.
- [ ] Regenerate all 109 module tasks and confirm exact coverage of the existing 245 proof-unit IDs.
- [ ] Confirm deterministic regeneration and byte-for-byte preservation of the current corpus.

Please keep this PR focused on generation and dataset artifacts. Checker, scoring, and runtime integration remain in #136.

src/dataset/proof_from_scratch/module_tasks.py
```diff
@@ -0,0 +1,735 @@
...
+    # Anything a dependency defines is handed to the agent for free, so left
+    # whole `Consensus.tla` gives away the `TypeOK`/`Inv` scaffolding a
+    # from-scratch task exists to make the agent rediscover. Seeded from every
+    # statement, and closed over the group because dependencies reference
+    # each other.
+    dep_paths = [path for _module, path in generate.layered_dep_paths(dump, str(source_root / spec_id), reachable)]
+    if dep_paths:
+        dep_dir = output_root / relative_dir / base_module
+        seeds = generate.referenced_identifiers(defs_text, model_text, *(text for _id, text in statements))
+        keep = generate.dep_keep_names(dep_paths, seeds, audit)
```
munimthahmid: I think the dependency pruning here is mixing names from the task with names from its dependency modules. For example, the local `Inv` in `Consensus/Voting.tla` causes the unrelated `Inv` from its `Consensus.tla` dependency to be kept, so the agent can access it as `C!Inv` even though that definition is not reachable from the selected statements. I found the same behavior for 13 definitions across 9 dependency files. Shouldn't the dependency closure keep the module identity instead of matching only the bare name?

sukanya1426: Yeah, you’re right. The keep set was using bare names globally, so Voting’s `Inv` also kept Consensus’s `Inv`, allowing the agent to read `C!Inv`. Fixed that, the closure now tracks `(module, name)` instead. Voting’s copied `Consensus.tla` now keeps `Init`/`Next`/`Spec` and drops `Inv`. I also regenerated the module suite (SANY-valid), added a homonym test, and pinned the shipped dependency. Thanksss....

src/dataset/proof_from_scratch/generate.py
```diff
@@ -585,13 +614,50 @@ def _scan_identifiers(text):
     out of a dependency that still needs it.
     """
     names = set()
-    for tok in _IDENTIFIER.findall(text):
+    for tok in _IDENTIFIER.findall(_text_without_strings(text)):
```
munimthahmid: I think qualified names are still being added to the bare-name seed set here. In `LockHS`, `P!Spec` keeps the unqualified `Spec` from `Lock.tla`, along with `Next` and `proc`, even though they aren't reachable from the selected statements. I tested excluding the qualified right-hand name from the bare seeds. It removed only those three definitions, and all 109 module tasks still passed SANY. Could you fix this case and add a regression test?

sukanya1426: Fixed it and added a regression test too.

sukanya1426: CC : @Qian-Cheng-nju I have completed the checklist, can you take a look ? Thanks.

Qian-Cheng-nju:
Thanks a lot for your hard work! Merged.

PR #138, opened 2026-08-30
Fix proof-from-scratch newline handling and CI verification

munimthahmid:
## What changed

I found and fixed two small issues in the proof-from-scratch pipeline.

- Extra EOF newlines were being rejected as scaffold changes and reported as cheating. They are now accepted, while other newline-only changes still fail without being marked as cheating.
- The generated module verifier already existed, but CI was not running it. CI now runs it on every PR.

## Testing

- The verifier regenerated and SANY-checked all 109 modules and 245 proof units with zero errors.
- A deliberately changed generated module was rejected with exit code 1.
- Added newline regression coverage and ran 114 focused tests. Ruff also passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,11 @@ jobs:
           TLAPS_BENCH_REQUIRE_PI_CLI: "1"
         run: uv run pytest tests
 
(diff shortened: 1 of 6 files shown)


Qian-Cheng-nju:
LGTM. Thanks!

PR #139, opened 2026-08-30
Route proof-from-scratch generate to the module-task suite.

sukanya1426:
### What changed
I found and fixed a leftover in the public generate command. After the module-task suite landed, tlaps-bench generate --mode proof-from-scratch was still generating the old per-theorem corpus under benchmark/proof-from-scratch/, while run uses benchmark/proof-from-scratch-module/.

So the documented command could overwrite the frozen 245-task source and wouldn’t generate the suite the evaluator uses. It now generates module tasks, with --output-dir / --source-dir mapped to the module-task flags. Rebuilding the 245 task corpus still uses uv run python -m dataset.proof_from_scratch.generate --layered.

### Testing
- Public CLI generated FindHighest as one module with 4 proof units, plus its Model/Defs files.
- tlaps-bench generate --mode proof-from-scratch --verify regenerated and verified all 109 modules / 245 proof units with 0 errors.
- Confirmed the existing 245-task tree is unchanged.
- Added CLI routing tests, Ruff passes.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -354,7 +354,7 @@ uv run tlaps-bench generate --mode proof-from-scratch
 uv run tlaps-bench generate --mode proof-from-scratch --verify
 ```
 
-Proof-from-scratch generation emits the module suite under `benchmark/proof-from-scratch-module/`: one task per source `spec_id`, with the existing 245 theorem IDs as proof units. It reads `benchmark/proof-from-scratch/` and will not write into that tree. `--verify` regenerates into a scratch directory and compares the shipped suite. To rebuild the frozen 245-task selection corpus itself, invoke `python -m dataset.proof_from_scratch.generate --layered` directly.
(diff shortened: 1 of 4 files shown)


sukanya1426:
@munimthahmid, Hi, can you please take a look at this? Is there anything else I should add here ?Thankss....

docs/USAGE.md
```diff
@@ -351,8 +351,11 @@ Regenerate benchmark files from annotated source specs.
 ```bash
 uv run tlaps-bench generate
 uv run tlaps-bench generate --mode proof-from-scratch
+uv run tlaps-bench generate --mode proof-from-scratch --verify
 ```
 
+Proof-from-scratch generation emits the module suite under `benchmark/proof-from-scratch-module/`: one task per source `spec_id`, with the existing 245 theorem IDs as proof units. It reads `benchmark/proof-from-scratch/` and will not write into that tree. `--verify` regenerates into a scratch directory and compares the shipped suite. To rebuild the frozen 245-task selection corpus itself, invoke `python -m dataset.proof_from_scratch.generate --layered` directly.
```
munimthahmid: This command fails from the repository root because `dataset` is under `src`. Could you change it to `uv run python -m dataset.proof_from_scratch.generate --layered` and make the same update in `design.md` and the PR description?

sukanya1426: Fixed

Qian-Cheng-nju:
LGTM! Thanks a lot!

PR #141, opened 2026-09-01
Fix benchmark result accounting

Qian-Cheng-nju:
## Summary

- distinguish scored timeouts from retriable grader and infrastructure errors
- aggregate specifications as PASS, FAIL, or UNRESOLVED without promoting excluded ERROR leaves
- resume interrupted logical attempts from saved proofs with cumulative accounting and only the original timeout remainder
- aggregate per-unit module obligations and report agent, grading, and total time separately
- preserve unavailable and lower-bound token usage instead of rendering it as zero

### DIFF
--- a/src/common/check_proof.py
+++ b/src/common/check_proof.py
@@ -1074,7 +1074,24 @@ def parse_strict_status(tlapm_exit, tlapm_output):
 
 
 MODULE_RESULT_PREFIX = "MODULE-RESULT: "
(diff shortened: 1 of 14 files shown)


Qian-Cheng-nju:
@sukanya1426 Could you take a look? Thanks!

sukanya1426 (approved):
Ran the scoring, resume, obligation, usage, and infra retry tests. Everything looks good on my side. Thanks.

PR #143, opened 2026-09-01
Remove approved-library duplicate from proof-from-scratch

### LINKED ISSUE #142: RFC: Policy for Proof-from-Scratch targets that match approved library theorems
## Context

Proof-from-Scratch agents can import modules from the frozen official proof-library catalog.

While auditing the new module suite, I found one target that exactly matches an approved library theorem.

The target `Cantor/Cantor10_NoSetContainsAllValues.tla` is:

```tla
THEOREM NoSetContainsAllValues ==
  \A S : \E x : x \notin S
```

The approved `TLAPS` module contains:

```tla
THEOREM NoSetContainsEverything ==
  \A S : \E x : x \notin S
```

The formulas are identical. Only the theorem names differ.

## Reproduction

This candidate changes only the permitted helper and proof regions:

```tla
LOCAL T == INSTANCE TLAPS

BY T!NoSetContainsEverything
```

The authoritative container checker accepts it:

```text
SANY-STATUS: valid
PROOF UNIT Cantor/Cantor10_NoSetContainsAllValues.tla: PASS
[INFO]: All 2 obligations proved.
PASS — 1/1 proof units are dependency-closed and trusted
```

The checker appears to follow the current official-library policy correctly. The question is whether this target still provides useful benchmark value.

The original source proof first proves `Cantor` and then derives this theorem as a corollary. The full source module passes 16 obligations with the current container TLAPM.

## Policy question

What is the desired policy for a target that is identical to a theorem in the approved library catalog?

Possible policies are:

1. Keep the target because finding and using an approved theorem is part of the task.
2. Remove the proof unit because it does not require proof construction after library discovery.
3. Replace it with another theorem from the source after a separate benchmark-value review.

The second option is the smallest dataset change. This proof unit is the only selected target in `Cantor/Cantor10.tla`.

This issue does not propose a task-specific library restriction. Such a restriction can make runs harder to compare and reproduce.

## Possible prevention

If exact library matches are not desired, CI can compare target formulas with approved library statements.

The comparison must use parsed formulas and normalized bound variables. It must run when the task selection or proof-library lock changes.

The current audit compared all 245 targets with 396 statements from the 37 approved modules. It found no other exact or alpha-equivalent match.

Saadmrp1038:
Closes #142


Qian-Cheng-nju:
LGTM. Thanks!

PR #144, opened 2026-09-12
Add MongoDB snapshot isolation proof task

Qian-Cheng-nju:
Add one Proof From Scratch task for MongoDB distributed transaction snapshot isolation:

```tla
(Spec /\ []SingleWritePerKey) => []SnapshotIsolation
```

The four upstream modules are copied unchanged from `74526c1201109405172eb845413154f547a815ee`, including the original `Index` definition. The wrapper formalizes the [maintainer's single-write-per-key assumption](https://github.com/mongodb-labs/vldb25-dist-txns/pull/3#issuecomment-5666002688) for transaction histories in `ops`, alongside the three snapshot-mode settings and `Timestamps \subseteq Nat`. It counts write occurrences by sequence position and permits writes to multiple keys and repeated reads. Timestamp zero is allowed; parameter sets have no cardinality bounds and timestamp values have no upper bound. No reference TLAPS proof is supplied.

Fix dependency discovery for parameterized instances such as `Storage(s)!Op`, retain the required storage definitions, and regenerate both task formats. Update provenance, attribution, and dataset tables. Existing tasks are unchanged; totals are 706 proof-completion tasks and 245 proof-from-scratch targets across 109 module tasks.

Update the SANY artifact lock to the September 14 release (`5dbdb42`), verified against its published SHA-256 digest.

Validation: [GitHub CI](https://github.com/specula-org/tlaps-bench/actions/runs/34937945526) passes. 256 local dataset tests, Ruff, and full generated-module verification pass. Exact staged task contexts pass SANY, and unfinished proofs receive no credit. Timestamp-domain controls reject Boolean and negative values and accept zero, large natural numbers, and the empty set. A separate TLAPS probe confirms that the generated assumptions support timestamp arithmetic. Bounded TLC checks complete 62,708 distinct states and 10,000 simulation traces, with repeated-write, multi-key, repeated-read, prepare-blocking, and late-commit controls. Details are in `source/MongoDB/README.md`.


will62794:
FYI independently proving correctness properties of the [Storage](https://github.com/mongodb-labs/vldb25-dist-txns/blob/main/Storage.tla) module could also be an interesting benchmark task. That module should independently satisfy snapshot isolation guarantees, perhaps along with some other, stronger properties that are relied on by the higher level distributed transactions protocol. I don't think we had formalized those properties exactly, but it may be an interesting problem as well to nail down those precise interface/contract invariants and prove them.

Qian-Cheng-nju:
That sounds like a great addition. Please feel free to share any formalization of the Storage module’s snapshot isolation guarantees or interface properties, even an initial draft would be helpful. I’d be happy to explore adding them to the benchmark. Thanks!

Issue #145, opened 2026-09-14
Serializable snapshot isolation benchmark


will62794:
The serializable snapshot isolation spec [here](https://github.com/pron/amazon-snapshot-spec/blob/master/serializableSnapshotIsolation.tla) could be another good benchmark. Proving the [CahillSerializable](https://github.com/pron/amazon-snapshot-spec/blob/9c60cb18151889d7b4c0a4ffd7de0b6fc2db0fb2/serializableSnapshotIsolation.tla#L1443-L1446) invariant would be one starting point. No inductive invariants or TLAPS proofs exist for this to my knowledge.

Qian-Cheng-nju:
Thank you very much for suggesting this! I’ve opened #146 to add `Spec => []CahillSerializable(history)` as a proof-from-scratch task. I made a few syntax adaptations for TLAPS compatibility, mainly converting recursive operators to recursive functions and expanding tuple binders, while retaining the original source alongside the adapted version. The PR is now under review.

PR #146, opened 2026-09-15
Add Cahill SSI proof-from-scratch task

Qian-Cheng-nju:
Adds the Cahill serializability task suggested in #145:

```tla
Spec => []CahillSerializable(history)
```

The specification is by Chris Newcombe for HPTS 2011, pinned to `pron/amazon-snapshot-spec@9c60cb18151889d7b4c0a4ffd7de0b6fc2db0fb2`. The wrapper adds no parameter bounds or behavioral assumptions. There is no reference proof.

The current TLAPM cannot parse the source's recursive operators and tuple binders. This PR therefore retains the original text, losslessly recoded to UTF-8, and includes a separate compatible rendering using recursive functions and scalar binders. The transformations are explicit and reproducible with `python3 source/CahillSSI/prepare.py --check`. Generated contexts remove both unproved source theorem declarations.

Validation:

- Original and compatible models each complete a two-key, two-transaction TLC check: 29,629 distinct states, including the target and Bernstein serializability checks. Both directions of refinement pass on that model; dependency graphs agree, and cycle helpers agree on all 512 directed graphs over three nodes.
- Each model also passes 2,000 three-transaction simulation traces. A write-skew control checks predicate rejection and protocol abort behavior. The generated task context passes the same 29,629-state target check.
- Flat and module tasks pass parsing and correctly reject an omitted proof. Full module verification reports zero errors; local tests report 2,052 passed and 2 skipped; Ruff passes.

These are bounded checks, not proofs of serializability or equivalence of the two representations. Source provenance, adaptation details, and validation scope are documented in `source/CahillSSI/README.md`.

The dataset now has 706 proof-completion tasks and 246 proof-from-scratch tasks, grouped into 110 module tasks for the latter mode.

Refs #145.


Qian-Cheng-nju:
@munimthahmid Could you review that? Thanks a lot!

munimthahmid (approved):
Everything looks good. I reran the TLC checks and verified the generated tasks, and everything passed. 

One small concern is that the upstream spec doesn’t have a license. But I think it's ok to adapt it as Chris is part of the team ?

